PR #3, opened 2021-06-25
[sailfishos][webrtc] Enable support for WebRTC video. JB#53982

d-grigorev:


rpm/0070-sailfishos-webrtc-Implement-video-capture-module.-JB.patch
```diff
@@ -0,0 +1,2237 @@
...
++#include "webrtc/base/refcount.h"
++#include "webrtc/base/scoped_ref_ptr.h"
++#include "webrtc/system_wrappers/include/trace.h"
++
++#include "webrtc/modules/video_capture/sfos/gst_droid_camera.h"
++
++namespace webrtc {
++namespace videocapturemodule {
++
++#define GST_DROID_CAMERA_PRIMARY   "primary"
```
abranson, line 493: Are these still valid? I think we have to use the camera number now

d-grigorev: Yes, this is reported as camera ID, see `GstDroidCameraManager::GetDeviceName()` below.

rpm/0070-sailfishos-webrtc-Implement-video-capture-module.-JB.patch
```diff
@@ -0,0 +1,2255 @@
...
++        uint32_t deviceUniqueIdUTF8Length,
++        char* productUniqueIdUTF8=0,
++        uint32_t productUniqueIdUTF8Length=0,
++        pid_t* pid=0);
++    bool QueryCapabilities (const char* deviceUniqueIdUTF8,
++                            std::vector<VideoCaptureCapability>& caps);
++    rtc::scoped_refptr<Camera> OpenCamera (const char* deviceUniqueIdUTF8);
++private:
++    int32_t FillCapabilities(int fd, std::vector<VideoCaptureCapability>& caps);
++    unsigned _numberOfDevices;
```
llewelld, line 1942: Is `_numberOfDevices` needed? It looks unused.

rpm/0070-sailfishos-webrtc-Implement-video-capture-module.-JB.patch
```diff
@@ -0,0 +1,2255 @@
...
++}
++
++int32_t VideoCaptureModuleSFOS::StopCapture()
++{
++    return _camera->StopCapture();
++}
++
++bool VideoCaptureModuleSFOS::CaptureStarted()
++{
++    return _camera->CaptureStarted();
```
llewelld, line 2048: Is it safe to assume `_camera != null` for these?

llewelld (commented):
I'm not able to do much more than a sanity check of the code I'm afraid. It broadly looks sensible and builds fine. I've made a few comments, but nothing major.

However, when I tried it on an XA2 with gmp-droid-0.2-1.2.1 I got some strange output when testing it with the [meetecho Janus test server](https://janus.conf.meetecho.com/videocalltest.html) you recommended. I get the same results with both the front and back cameras, but I don't know why this is I'm afraid.

rpm/0069-sailfishos-webrtc-Enable-GMP-for-encoding.-JB-53982.patch
```diff
@@ -0,0 +1,477 @@
...
+-  virtual uint32_t NumInUse(GMPSharedMem::GMPMemoryClasses aClass);
++  virtual int32_t NumInUse(GMPSharedMem::GMPMemoryClasses aClass);
+ 
+   // These have to be implemented using the AllocShmem/etc provided by the
+   // IPDL-generated interfaces, so have the Parent/Child implement them.
+@@ -70,6 +71,7 @@ class GMPSharedMemManager {
+                      ipc::Shmem* aMem) = 0;
+   virtual void Dealloc(ipc::Shmem& aMem) = 0;
+ 
++  Mutex mMutex;
```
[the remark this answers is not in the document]
d-grigorev: I accidentally left it public. Thank you.

rpm/0069-sailfishos-webrtc-Enable-GMP-for-encoding.-JB-53982.patch
```diff
@@ -0,0 +1,477 @@
...
+index 1fa2425a3c2c..6bfde6b04e26 100644
+--- a/dom/media/gmp/GMPVideoDecoderParent.cpp
++++ b/dom/media/gmp/GMPVideoDecoderParent.cpp
+@@ -134,9 +134,9 @@ nsresult GMPVideoDecoderParent::Decode(
+   // Very rough kill-switch if the plugin stops processing.  If it's merely
+   // hung and continues, we'll come back to life eventually.
+   // 3* is because we're using 3 buffers per frame for i420 data for now.
+-  if ((NumInUse(GMPSharedMem::kGMPFrameData) >
+-       3 * GMPSharedMem::kGMPBufLimit) ||
++  if ((NumInUse(GMPSharedMem::kGMPFrameData) > 3 * GMPSharedMem::kGMPBufLimit) ||
```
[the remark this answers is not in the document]
d-grigorev: Yes, I've been fiddling with these lines. This is a candidate for future revision. Will remove this from the patch.

rpm/0069-sailfishos-webrtc-Enable-GMP-for-encoding.-JB-53982.patch
```diff
@@ -0,0 +1,477 @@
...
++      return nullptr;
++    }
++
++    if (HaveGMPFor(NS_LITERAL_CSTRING(GMP_API_VIDEO_DECODER), { tag })) {
++      decoder = GmpVideoCodec::CreateDecoder(aType);
+       if (decoder) {
+         mRecvCodecPlugin = static_cast<WebrtcVideoDecoder*>(decoder);
+       }
+-      break;
++      return decoder;
```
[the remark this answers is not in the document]
d-grigorev: I did this on purpose. If `HaveGMPFor()` reported codec availability, then null indicates an error when creating a GMP codec. Otherwise it may hide problems. Do you think it's wrong?

rpm/0069-sailfishos-webrtc-Enable-GMP-for-encoding.-JB-53982.patch
```diff
@@ -0,0 +1,477 @@
...
+@@ -695,11 +738,29 @@ int32_t WebrtcGmpVideoDecoder::GmpInitDone(GMPVideoDecoderProxy* aGMP,
+   GMPVideoCodec codec;
+   memset(&codec, 0, sizeof(codec));
+   codec.mGMPApiVersion = 33;
++  codec.mWidth = mCodecSettings.width;
++  codec.mHeight = mCodecSettings.height;
+ 
+-  // XXX this is currently a hack
+-  // GMPVideoCodecUnion codecSpecific;
+-  // memset(&codecSpecific, 0, sizeof(codecSpecific));
```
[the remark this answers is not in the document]
d-grigorev: Yes. it is not relevant for gmp-droid.

rpm/0070-sailfishos-webrtc-Implement-video-capture-module.-JB.patch
```diff
@@ -0,0 +1,2255 @@
...
++
++    close(fd);
++
++    char cameraName[64];
++    memset(deviceNameUTF8, 0, deviceNameLength);
++    memcpy(cameraName, cap.card, sizeof(cap.card));
++
++    if (deviceNameLength >= strlen(cameraName))
++    {
++        memcpy(deviceNameUTF8, cameraName, strlen(cameraName));
```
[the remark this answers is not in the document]
d-grigorev: Yeah. I copied this code from `media/webrtc/trunk/webrtc/modules/video_capture/linux/` and edited just to fit my internal Camera interface. I never tested it and it is not listed in supported camera types. Will review that better.

rpm/0070-sailfishos-webrtc-Implement-video-capture-module.-JB.patch
d-grigorev: Yes. The module will not be created it it has not passed `Init()` in `VideoCaptureImpl::Create()`.

rpm/0069-sailfishos-webrtc-Enable-GMP-for-encoding.-JB-53982.patch
```diff
@@ -0,0 +1,477 @@
...
+-  // XXX At this point, we only will get mode1 data (a single length and a
+-  // buffer) Session_info.cc/etc code needs to change to support mode 0.
+-  *(reinterpret_cast<uint32_t*>(frame->Buffer())) = frame->Size();
+-
+-  // XXX It'd be wonderful not to have to memcpy the encoded data!
+-  memcpy(frame->Buffer() + 4, aDecodeData->mImage._buffer + 4,
+-         frame->Size() - 4);
++  // gmp-droid does not require conversion of NAL units for h264
++  // TODO: Check OpenH264 requirements
++  memcpy(frame->Buffer(), aDecodeData->mImage._buffer, frame->Size());
```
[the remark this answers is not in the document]
d-grigorev: Yes. This copies the encoded data. There was a conversion for h264, but now it's not needed. I will rephrase the comment to make it clear.

d-grigorev:
> I'm not able to do much more than a sanity check of the code I'm afraid. It broadly looks sensible and builds fine. I've made a few comments, but nothing major.
> 
> However, when I tried it on an XA2 with gmp-droid-0.2-1.2.1 I got some strange output when testing it with the [meetecho Janus test server](https://janus.conf.meetecho.com/videocalltest.html) you recommended. I get the same results with both the front and back cameras, but I don't know why this is I'm afraid.
> 
> 

That's bad. Looks like the image format from viewfinder is totally different. I have no access to XA2 to check, and gecko is cut off the logging in production build. I will ask is someone has the device.

rpm/0069-sailfishos-webrtc-Enable-GMP-for-encoding.-JB-53982.patch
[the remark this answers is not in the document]
llewelld: That makes perfect sense and I was just checking. Thanks for the explanation.

rpm/0069-sailfishos-webrtc-Enable-GMP-for-encoding.-JB-53982.patch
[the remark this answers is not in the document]
llewelld: Thanks; again just checking.

rpm/0070-sailfishos-webrtc-Implement-video-capture-module.-JB.patch
llewelld: Thanks for clarifying.

rpm/0070-sailfishos-webrtc-Implement-video-capture-module.-JB.patch
```diff
@@ -0,0 +1,2255 @@
...
++    _currentCaptureSettings.rawType = kVideoUnknown;
++    _currentCaptureSettings.width = capability.width;
++    _currentCaptureSettings.height = capability.height;
++    _currentCaptureSettings.maxFPS = capability.maxFPS;
++    /* FIXME: Is it possible to read capture delay from the HW? */
++    _currentCaptureSettings.expectedCaptureDelay =
++            mozilla::Preferences::GetUint(EXPECTED_CAPTURE_DELAY_PREF, 500);
++
++    /* Pre-calculate frame size for YV12 */
++    _frameSize = capability.width * capability.height * 3 / 2;
```
adenexter, line 674: This size may be too small if the camera source adds some padding to the row width for memory alignment reasons.

gst-droid calculates its buffer sizes here https://github.com/sailfishos/gst-droid/blob/master/gst-libs/gst/droid/gstdroidmediabuffer.c#L260. I'm not sure how much of the info can be extracted from gstreamer ahead of time, but it should become available when you map the gstreamer buffer.

rpm/0070-sailfishos-webrtc-Implement-video-capture-module.-JB.patch
```diff
@@ -0,0 +1,2255 @@
...
++    g_main_context_push_thread_default (_captureContext);
++
++    /* TODO: Make the pipeline configurable */
++    _pipeline = gst_parse_launch (
++        "droidcamsrc mode=2 exposure-mode=auto iso_speed=0 name=src"
++        " src.vfsrc"
++        " ! capsfilter name=camera_capsfilter"
++        " ! videorate drop-only=true"
++        " ! capsfilter name=rate_capsfilter"
++        " ! queue ! appsink name=videosource", NULL );
```
adenexter, line 532: I realise build repo dependencies make this impossible at the moment, but if it could be resolved so that droidmedia is available to use directly would that be an option?

It's not necessarily going to be any easier to use, but it may open up some option to supply this API's buffer type to the camera API and eliminate a copy.

d-grigorev: The gecko has some support for native buffers, so (after a massive patching and rewriting) it can pass native buffers directly to the encoder. It's my goal for the future. I tried to create a capture module based on plain droidmedia: although it calls nearly the same droidmedia functions as its gst version, I've been unable to get it to produce buffers. Perhaps I'm missing something.

adenexter: What do you mean by native buffers here? Because the android camera has two optimized buffer types it outputs. There are the preview stream graphics memory buffers (gralloc) which can be consumed by the GPU directly without any memory  copies. And there are the recording stream metadata headers which can be consumed by the hardware accelerated video encoders. And just to complicate things further there is a way to connect the recording stream to the video encoder in the android media server so the application doesn't see the video data until after it is encoded, this is what gst-droid utilizes currently and although there is an option to do the data passing in the application my understanding is that it doesn't currently and perhaps never worked.

I don't think there's any trick to using droidmedia directly at least as far as getting preview frames is concerned. It should be a matter of creating the camera object, locking it, setting up callbacks on both the camera and its buffer queue, then starting the preview stream. Then graphics buffers should start arriving in the buffer queue callbacks. There are also some flags you can set to get raw data in the preview_frame_cb callback, although that's only valid until the callback returns so there's a necessary copy to work with that most likely.

This is what I've tried as a standalone thing https://gist.github.com/adenexter/89375f20793fcee698d623597b84ff0e. It's not displaying anything but you can at least see it's querying the cameras and starts one and is continually receiving frames. 

On a Jolla C, and a T8 it's producing both preview and recording frames, but on an XA2 the recording stream doesn't work and as seen in David's screenshot the color format of the unlocked graphics buffers is not YV12 as assumed but is instead HAL_PIXEL_FORMAT_IMPLEMENTATION_DEFINED. In the parameters the preview-format is listed as yuv420sp as it is on the other devices but at I guess only the data in the preview_frame_cb callback is in that format. It can probably be fixed in the gstreamer implementation by dropping the memory:DroidMediaQueueBuffer feature from the caps negotiation, probably at the cost of another copy.

d-grigorev: By native handle I mean graphics memory buffers. I was thinking about passing a handle to such buffer to the GMP encoder plugin instead of managing shared memory and copying data back and fourth.

Thank you for giving the example of using droidmedia. Your code works, but I still can't see the difference with mine. Interesting, will look closer at this.

I was able to repeat David's result at my Sony 10. I'm using viewfinder frames in my capture code. Can't prove it but I think the data is in RGB format and perhaps that's due to the buffer's usage [here](https://github.com/sailfishos/droidmedia/blob/766eb45a873dee3941eb4d455a92544eddf78c4f/private.cpp#L86): 

m_queue->setConsumerUsageBits(android::GraphicBuffer::USAGE_HW_TEXTURE);

As far as I understand, to make it YUV it's enough to set it to USAGE_HW_VIDEO_ENCODER, and perhaps that's why recorder mode doesn't work for some devices, what do you think? I don't have access to adaptation repo's so I can't build a hacked version, unfortunately.

adenexter: > As far as I understand, to make it YUV it's enough to set it to USAGE_HW_VIDEO_ENCODER, and perhaps that's why recorder mode doesn't work for some devices, what do you think? I don't have access to adaptation repo so I can't build a hacked version, unfortunately.

I can try it, but I'm not expecting different results. I think the usage bits are more hints to the driver about what memory allocation and locking strategies might work best for the usage and not all hints will apply to all buffer producers. So far as I understand the graphics buffers aren't really intended to be used for read back, that's what the data in the CAMERA_MSG_PREVIEW_FRAME and CAMERA_MSG_VIDEO_FRAME messages are for. It's unfortunate that droidmedia is abstracting away the reference counted lifetime of the data from the CAMERA_MSG_PREVIEW_FRAME  message.

The recording failure on the XA2 is probably a bug in the implementation for later android versions. As it's not used by gst-droid it has quite likely gone untested.

adenexter: So looking at logcat output the video recording problem in more recent android versions on some hardware at least is that only the  [`VIDEO_BUFFER_MODE_BUFFER_QUEUE`](https://android.googlesource.com/platform/frameworks/av/+/master/camera/include/camera/android/hardware/ICamera.h#48) mode is supported

`07-07 01:32:40.307  3076 23883 E Camera2Client: startRecordingL: Camera 0: Recording only supported buffer queue mode, but mode 0 is requested!`

Which can be enabled by calling `droid_media_camera_store_meta_data_in_buffers(.., true)` But then recording still fails because that requires that a [buffer queue be set as the video target](https://android.googlesource.com/platform/frameworks/av/+/master/camera/include/camera/android/hardware/ICamera.h#141) and nothing does that. It also means my understanding of the architecture's somewhat out of date.

d-grigorev: Thank you for trying this. Looks like the viewfinder frames on Sony devices are in some kind of proprietary format, so nothing left except to try to implement a video target for recorder mode.

adenexter: Changing the queue usage flag had no effect.  I also tried setting a queue as the video target, but that's still failing but now on an unsupported format error followed by some sort of lock up. I'm not sure what's happening there, the format is being left as the default which in the parameters is `video-frame-format=android-opaque` and changing that doesn't appear to have any effect.

Now that I've gone through the process myself; the repo manifests for the Sony device adaptions are in https://github.com/mer-hybris/android so everything you need to build it yourself should be publicly available.

d-grigorev: Thank you. I have exactly the same result on MediaTek device. For some unknown reason HAL chose format to be RGBA_8888, but then it reports that a scaler is missing for this format. Changing usage to `USAGE_HW_VIDEO_ENCODER` didn't help (however [here](http://androidxref.com/9.0.0_r3/xref/device/generic/goldfish/camera/EmulatedQemuCamera3.cpp#887) it has a key role), so I keep diving into Android code.

d-grigorev: It seems that the only way to get raw frames in a platform independent way is to use HAL_PIXEL_FORMAT_YCBCR_420_888 for the buffer queue and map buffers using [lockYCbCr](http://androidxref.com/9.0.0_r3/xref/frameworks/native/include/ui/GraphicBuffer.h#168). I updated my PR https://github.com/sailfishos/droidmedia/pull/81. I think it's too complex to add a proper support for this to gst-droid, so the next step is rewriting the capture code using plain droidmedia. As far as I understand, adding droidmedia headers to pj:oss is non-trivial, so the code will be split into a device-specific plugin and the plugin API will be hard-coded into gecko.

adenexter: There are definite advantages to having a plugin API, shorter build times if you're just making changes to it being my favorite.

But for the record, a hardware adaptation independent droidmedia-devel package should in theory be almost trivial. The issues I can identify are; You'd have to create a separate .spec file that manually installs the headers instead of using the Android makefile. There is an ifdef in the droidmedia constants header file that renames some members of a struct to follow name changes between Android versions which could be replaced with a union or just picking one of the names. And instead of linking against libhybris-common you would have to dlopen it and manually resolve the android_dlopen and android_dlsym symbols.

It's not a necessity, but there is a developing pattern of project specific plugins to utilize the glue code for loading for what is effectively a droidmedia plugin. If short cutting that is the same or less work than a new plugin API then I'd consider it.

adenexter: I was thinking specifically of gst-droid and nemo-gst-interfaces. Because the devel package of droidmedia is built per device, that means gst-droid is per device and can't provide its own devel package that applications can use. So we have this other package nemo-gst-interfaces that is device agnostic but exclusively contains gst-droid specific interfaces interfaces to act as a go between. And if you ever have reason to change one of those interfaces it's a three project minimum edit. And then we also have a plugin for qtmultimedia to act as an installable bridge between it and nemo-gst-interfaces.

That last point is something I overlooked earlier, not all devices will have droidmedia installed, and selectively installing a plugin for the browser is probably the best way to manage that. I'd still like to be able to build that plugin and gst-droid without a hardware specific sdk though.

d-grigorev:
Although I tested it on a sony device and it works, I found out that sony's video encoder doesn't support I420 pixel format input. The problem is in gmp-droid, a fix PR is coming soon at the cost of another copy of video data. Unfortunately sony uses semi-planar format internally, which is not compatible with gecko's internals. Gecko has surprisingly bad design of video input, however I see some optimization opportunities here.
The only known working scenario is local echo, like https://mozilla.github.io/webrtc-landing/gum_test.html.

d-grigorev:
This patch uses https://github.com/d-grigorev/amber-camera/tree/initial as a video input library. It contains two specs, a library and a plugin. The plugin uses droidmedia-devel at build time and droidmedia at runtime.

d-grigorev:
The encoder issue is fixed in https://github.com/sailfishos/gmp-droid/pull/15.

rpm/0071-sailfishos-webrtc-Implement-video-capture-module.-JB.patch
```diff
@@ -0,0 +1,797 @@
...
+-        ldflags += [
+-          "-lrt",
+-          "-lXext",
+-          "-lX11",
+-        ]
+-      }
+ 
+       deps = [
+         ":video_capture_internal_impl",
+diff --git a/media/webrtc/trunk/webrtc/modules/video_capture/linux/device_info_linux.cc b/media/webrtc/trunk/webrtc/modules/video_capture/linux/device_info_linux.cc
```
adenexter, line 193: This doesn't apply because you dropped the patch it's reverting.

rpm/0071-sailfishos-webrtc-Implement-video-capture-module.-JB.patch
```diff
@@ -0,0 +1,797 @@
...
+-        "linux/device_info_linux.cc",
+-        "linux/device_info_linux.h",
+-        "linux/video_capture_linux.cc",
+-        "linux/video_capture_linux.h",
++        "gst/device_info_gst.cc",
++        "gst/device_info_gst.h",
++        "gst/video_capture_gst.cc",
++        "gst/video_capture_gst.h",
++        "gst/gst_camera.cpp",
++        "gst/gst_camera.h",
```
adenexter, line 175: Gstreamer's not used for anything now right? Should these be removed or replaced with sfos equivalents?

rpm/0071-sailfishos-webrtc-Implement-video-capture-module.-JB.patch
```diff
@@ -0,0 +1,797 @@
...
++#include <unistd.h>
++
++#include "mozilla/Preferences.h"
++#include "webrtc/system_wrappers/include/logging.h"
++
++#include "webrtc/modules/video_capture/sfos/device_info_sfos.h"
++
++#define EXPECTED_CAPTURE_DELAY_PREF "media.getusermedia.camera.expected_capture_delay"
++#define DROID_CAMERA_PRIMARY    "primary"
++#define DROID_CAMERA_SECONDARY  "secondary"
```
adenexter, line 245: Are these used anywhere?

rpm/0071-sailfishos-webrtc-Implement-video-capture-module.-JB.patch
```diff
@@ -0,0 +1,797 @@
...
++        LOG_F(LS_ERROR) << "Device name too long";
++        return -1;
++    }
++
++    FillCapabilities(deviceUniqueIdUTF8);
++
++    // Store the new used device name. The parent class needs this for some reason
++    _lastUsedDeviceNameLength = deviceUniqueIdUTF8Length;
++    _lastUsedDeviceName = (char*) realloc(_lastUsedDeviceName,
++                                                   _lastUsedDeviceNameLength + 1);
```
adenexter, line 336: Could this be a std::string?

rpm/0071-sailfishos-webrtc-Implement-video-capture-module.-JB.patch
```diff
@@ -0,0 +1,797 @@
...
++    LOG_F(LS_INFO) << "VideoCaptureModuleSFOS::Notify ScreenConfiguration.orientation: " << aConfiguration.orientation();
++    _screenRotationAngle = ScreenOrientationToAngle(aConfiguration.orientation());
++    UpdateCaptureRotation();
++}
++
++int VideoCaptureModuleSFOS::ScreenOrientationToAngle(mozilla::dom::ScreenOrientationInternal orientation)
++{
++    switch (orientation) {
++    /* FIXME: What is the default orientation? */
++    case mozilla::dom::eScreenOrientation_Default:
```
adenexter, line 606: Sailfish devices always default to portrait. The SDK emulator might be an exception to that, but that might only be an issue at the compositor level.

adenexter (changes requested):
Once I resolved the issue with applying the last patch the echo test worked on an Xperia XA2 which was one of the earlier problem devices.

Not a lot stands out in the code otherwise.

I guess we'll need to sort out a process for promoting and reviewing amber-camera as well. The standout thing to me is there might be some similar push back to the separate spec file for the droidmedia plugin. It might be better received if it's in a separate project, or combined into a single spec with conditions.

d-grigorev:
> I guess we'll need to sort out a process for promoting and reviewing amber-camera as well.

The layers gave permission to contibute it to github.com/sailfishos, so what are next steps?. You can leave comments in https://github.com/d-grigorev/amber-camera/pull/1.

> It might be better received if it's in a separate project, or combined into a single spec with conditions.

Isn't that convenient to have multiple specs? amber-camera headers should be accessible from pj:oss, amber-camera-droid-plugin should be built in a (common) adaptation project. The spec will be chosen depending on the OBS project name - looks clear for me. What are the disadvantages?

adenexter:
> 
> 
> > I guess we'll need to sort out a process for promoting and reviewing amber-camera as well.
> 
> The layers gave permission to contibute it to github.com/sailfishos, so what are next steps?. You can leave comments in [d-grigorev/amber-camera#1](https://github.com/d-grigorev/amber-camera/pull/1).

An admin will need to create the project under the sailfishos group. Could you create a task in bugzilla requesting it?

> > It might be better received if it's in a separate project, or combined into a single spec with conditions.
> 
> Isn't that convenient to have multiple specs?

Refer to the discussion in the droidmedia split bug. I argued for separate files there because droidmedia in a unique project, but based on that I would expect a similar objection again.

rpm/0071-sailfishos-webrtc-Implement-video-capture-module.-JB.patch
d-grigorev: This is the member of the parent class. Other capture implementations also fill it. Do you think I should replace it everywhere to std::string? I'd happy to, but it will make porting to future versions of gecko a bit harder.

adenexter: No, don't do that. I should have found the definition.

d-grigorev:
Changes from the previous version:
* The patch now applies cleanly
* Fixed reported issues
* Fixed hardware-backed VP8 encoding using gmp-droid

rpm/0071-sailfishos-webrtc-Implement-video-capture-module.-JB.patch
d-grigorev: Thank you. I changed the comment.

rpm/0071-sailfishos-webrtc-Implement-video-capture-module.-JB.patch
d-grigorev: These are leftovers from previous implementation. Removed, thank you.

d-grigorev:
Changes from the previous version:
* Change the name of the video input library to gecko-camera, see https://github.com/d-grigorev/gecko-camera/pull/1

adenexter (commented):
I still want to verify this works on a Jolla C and I checked in on the promotion build of the droidmedia changes this depends on and there's a couple compilation failures related to age of the build environment. 

But if I can see this working there, I think I'm ready to approve.

d-grigorev:
> I still want to verify this works on a Jolla C and I checked in on the promotion build of the droidmedia changes this depends on and there's a couple compilation failures related to age of the build environment.

Unfortunately I don't have a Jolla C. The code for YCbCr template may need to be fixed in gecko-camera, if the picture is weird.

> But if I can see this working there, I think I'm ready to approve.

Thank you. I have a fix for audio getting out of sync with video in the incoming stream, but I think it is matter of the following commits.

adenexter:
> Unfortunately I don't have a Jolla C. The code for YCbCr template may need to be fixed in gecko-camera, if the picture is weird.
> 
The droidmedia promotion has succeeded on that hardware now. And it appears to work without any display issues.

adenexter (approved):

okodron (approved):
Looks like it's ready now.

PR #6, opened 2021-07-15
[xulrunner-qt5] Fix textarea content being clipped when the keyboard is visible. Fixes JB#54847 OMP#JOLLA-244

adenexter:
As PrepareViewport() does for the global transform, disable Y coordinate
flipping of scissor rectangles when targeting an offscreen window
surface.


llewelld (approved):
I tested using the OBS builds and it fixed the problem nicely. So this looks good to me. I've two suggestions which you're welcome to ignore (I can see reasons why you might not want to apply them):
1. It looks like both changes could be simplified to use `FlipY(<rect>.YMost())` in their final clauses.
2. Would it make sense to combine patches 65 and 70?

adenexter:
>     1. It looks like both changes could be simplified to use `FlipY(<rect>.YMost())` in their final clauses.

Done

>     2. Would it make sense to combine patches 65 and 70?

Aside from fewer patches is better, nah. I mean there is some probably some original change which if done without error would have included both changes, but now you're describing two different issues in the change, and the PR becomes a diff of a diff.

llewelld (approved):
Thanks for the changes and explanation. Just re-confirming my approval.

paveltkv (approved):

PR #7, opened 2021-07-24
Adapt code from EmbedLiteViewBaseChild/Parent to EmbedLiteViewChild/Parent. JB#54983 OMP#JOLLA-261

llewelld:
The `EmbedLiteViewBaseChild`/`Parent` classes previously inherited by `EmbedLiteViewChild`/`Parent` have now been integrated into `EmbedLiteViewChild`/`Parent`. This shifts various changes made to the former into the latter, specifically to support:

1. Setting desktop mode.
2. Setting a customer HTTP user agent.
3. Handling the virtual keyboard margin.


llewelld:
This PR is just moving some code around and updating it for the esr78 interfaces, but not adding any new functionality

chriadam:
Thank you very much.  The alternative would be to modify the cherry-picks in my branch to move this code properly (clearly when I did the cherry-picks, the way I resolved the conflicts was very suboptimal, and I didn't notice at the time as I didn't have a working build environment).

So, I have no preference, we can either: 1) cherry pick this as-is into my branch, and use that one going forward, or 2) throw away my branch, re-do the cherry-picks with your changes included in each case.

chriadam:
Obsoleted by https://github.com/sailfishos/gecko-dev/pull/11

PR #8, opened 2021-07-30
[sailfishos][gecko] Patch rust components to build on arm. JB#54221

llewelld:
libloading mixes both rust and C code. When build on for an arm target
using sb2 the C code in global_static.c ends up in elf32-littleendian
format, whereas it should be elf32-i486.

This patch overrides the default build step using cc::Build to force the
right tool to be used (host-gcc) and allow libloading to be compiled and
linked against.


Thaodan:
>This patch overrides the default build step using cc::Build to force the
>right tool to be used (host-gcc) and allow libloading to be compiled and
>linked against.

llewelld:
> Should this maybe be the commit message

I should have made this clearer, but the intention isn't to merge this change; it's only to demonstrate the problem. I'm hoping there's a more general solution for `sb2` that means this won't be necessary.

But I completely agree with your comment, and if it looks like there's any chance of this getting merged, I'll make sure I fix the commit message.

llewelld:
This has been superseded by #20 and is no longer needed, so closing.

PR #2, opened 2021-06-25
[sailfishos][gecko][rpm] Use system sqlite. Fixes JB#54800

paveltkv:

PR #13, opened 2021-08-13
Initialize profiler properly

rainemak:
This PR contains two main fixes. In addition, there are two small cleanup commits.

The first commit fixes initialization sequence. It should be following and there should
be reverse order in the TermEmbedding.

- NS_LogInit - sets main thread
- IOInterposer::Init
- profiler_init

GeckoProfiler.h states that profiler_init call must happen before any
other profiler calls.

The second commit removes wrong NS_LogTerm call as we do not have
"event loop" per se rather the constructed worker thread is our gecko main thread. Hence,
NS_LogTerm call should be only called from the TermEmbedding.

### DIFF
--- a/embedding/embedlite/EmbedLiteApp.cpp
+++ b/embedding/embedlite/EmbedLiteApp.cpp
@@ -32,8 +32,6 @@
 #include "EmbedLiteCompositorBridgeParent.h"
 #include "EmbedLiteAppProcessParent.h"
 
-#include "GeckoProfiler.h"
-
 namespace mozilla {
 namespace startup {
 extern bool sIsEmbedlite;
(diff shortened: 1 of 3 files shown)


okodron (approved):

paveltkv (approved):

rainemak:
Rebase

chriadam (approved):
LGTM.

With this, I still get the mozilla::net::CookiePersistentStorage::TryInitDB() segfault on line:

```
rv = mStorageService->OpenUnsharedDatabase(mCookieFile,
                                               getter_AddRefs(mSyncConn));
```
but that's a separate issue, not profiler related.

So, LGTM.

rainemak:
> LGTM.
> 
> With this, I still get the mozilla::net::CookiePersistentStorage::TryInitDB() segfault on line:
> 
> ```
> rv = mStorageService->OpenUnsharedDatabase(mCookieFile,
>                                                getter_AddRefs(mSyncConn));
> ```
> 
> but that's a separate issue, not profiler related.
> 
> So, LGTM.

Try to cleanup .local/share/org.sailfishos/browser/

Further, it might well be that we need to bring back https://github.com/sailfishos/gecko-dev/pull/2 but let's do not rush to conclusions.

PR #20, opened 2021-08-19
Apply build process change for arm build

llewelld:
Various changes needed to get the armv7hl build to pass through.

See the individual commit comments for the details.

### DIFF
--- a/rpm/0033-sailfishos-configure-Read-rustc-host-from-environmen.patch
+++ b/rpm/0033-sailfishos-configure-Read-rustc-host-from-environmen.patch
@@ -1,0 +1,91 @@
+From 637a3be1123efa220a712de3f8d697b8c648d0d1 Mon Sep 17 00:00:00 2001
+From: Raine Makelainen <raine.makelainen@jolla.com>
+Date: Mon, 28 Jun 2021 12:52:10 +0300
+Subject: [PATCH] [gecko][configure] Read rustc host from environment. JB#53019
+ OMP#JOLLA-93
+
+Signed-off-by: Raine Makelainen <raine.makelainen@jolla.com>
+---
+ build/moz.configure/rust.configure       |  3 ++-
+ third_party/rust/cc/.cargo-checksum.json |  2 +-
+ third_party/rust/cc/src/lib.rs           | 16 ++++++++++------
+ 3 files changed, 13 insertions(+), 8 deletions(-)
+
+diff --git a/build/moz.configure/rust.configure b/build/moz.configure/rust.configure
+index e5122d6b86e9..20a2058f30d8 100644
+--- a/build/moz.configure/rust.configure
++++ b/build/moz.configure/rust.configure
+@@ -91,6 +91,7 @@ set_config('RUSTC', rustc)
+ 
+ @depends_if(rustc)
+ @checking('rustc version', lambda info: info.version)
++@imports('os')
+ def rustc_info(rustc):
+     if not rustc:
+         return
+@@ -99,7 +100,7 @@ def rustc_info(rustc):
+     return namespace(
+         version=Version(info.get('release', '0')),
+         commit=info.get('commit-hash', 'unknown'),
+-        host=info['host'],
++        host=os.environ.get('RUST_HOST_TARGET'),
+         llvm_version=Version(info.get('LLVM version', '0')),
+     )
+ 
+diff --git a/third_party/rust/cc/.cargo-checksum.json b/third_party/rust/cc/.cargo-checksum.json
+index 417fde7953fd..72795ff19b70 100644
+--- a/third_party/rust/cc/.cargo-checksum.json
++++ b/third_party/rust/cc/.cargo-checksum.json
+@@ -1 +1 @@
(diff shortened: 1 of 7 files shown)


rpm/xulrunner-qt5.spec
```diff
@@ -356,13 +418,13 @@ echo "ac_add_options --disable-elf-hack" >> "$MOZCONFIG"
...
 # Additionally we limit the memory usage during linking
- echo 'FIX_LDFLAGS="-Wl,--reduce-memory-overheads -Wl,--no-keep-memory -Wl,-rpath=%{mozappdir}"' >> "${MOZCONFIG}"
+ echo 'FIX_LDFLAGS="-Wl,--strip-debug -Wl,--gc-sections -Wl,--reduce-memory-overheads -Wl,--no-keep-memory -Wl,-rpath=%{mozappdir}"' >> "${MOZCONFIG}"
  echo 'export LDFLAGS="$FIX_LDFLAGS"' >> "${MOZCONFIG}"
  echo 'LDFLAGS="$FIX_LDFLAGS"' >> "${MOZCONFIG}"
  echo 'export WRAP_LDFLAGS="$FIX_LDFLAGS"' >> "${MOZCONFIG}"
  echo 'mk_add_options LDFLAGS="$FIX_LDFLAGS"' >> "${MOZCONFIG}"
 
-RPM_BUILD_NCPUS=`nproc`
+RPM_BUILD_NCPUS=4
```
xfade, line 427: Not sure if that is still needed anymore? Was this a debug thing?

rpm/xulrunner-qt5.spec
```diff
@@ -277,10 +304,45 @@ export SB2_RUST_USE_REAL_EXECVP=Yes
...
+
+#%ifarch %arm
+echo "ac_add_options --host=armv7-unknown-linux-gnueabihf" >> "$MOZCONFIG"
+echo "ac_add_options --target=armv7-unknown-linux-gnueabihf" >> "$MOZCONFIG"
+#%endif
+
 # hack for when not using virtualenv
 ln -sf "%BUILD_DIR"/config.status $PWD/build/config.status
 
 %ifarch %arm
```
xfade, line 337: Needs aarch64 too

rpm/xulrunner-qt5.spec
```diff
@@ -254,20 +272,29 @@ echo "export QT_QPA_PLATFORM=minimal" >> "%BUILD_DIR"/rpm-shared.env
...
 # for native x86 builds
+echo "export SB2_RUST_TARGET_TRIPLE=%SB2_TARGET" >> "%BUILD_DIR"/rpm-shared.env
+echo "export RUST_HOST_TARGET=%SB2_TARGET" >> "%BUILD_DIR"/rpm-shared.env
+
+echo "export RUST_TARGET=%SB2_TARGET" >> "%BUILD_DIR"/rpm-shared.env
+echo "export TARGET=%SB2_TARGET" >> "%BUILD_DIR"/rpm-shared.env
+echo "export HOST=%SB2_TARGET" >> "%BUILD_DIR"/rpm-shared.env
+echo "export SB2_TARGET=%SB2_TARGET" >> "%BUILD_DIR"/rpm-shared.env
+
 %ifarch %arm
```
xfade, line 288: Add aarch64 here too as it got stripped out below?

rpm/0033-sailfishos-configure-Read-rustc-host-from-environmen.patch
```diff
@@ -0,0 +1,109 @@
...
+         let flags_env_var_value = self.get_var(flags_env_var_name);
+         if let Ok(_) = flags_env_var_value {
+             true
+@@ -1896,7 +1897,8 @@ impl Build {
+         let default = if host.contains("solaris") {
+             gnu
+         } else {
+-            traditional
++            //traditional
++            "armv7-unknown-linux-gnueabihf"
```
xfade, line 61: Are we sure this works when building for aarch64?

rpm/xulrunner-qt5.spec
Thaodan: or better %arm64

rpm/xulrunner-qt5.spec
Thaodan: or better %arm64 and change %arm to %arm32 (see: https://github.com/rpm-software-management/rpm/blob/rpm-4.16.x/macros.in#L1137).

Thaodan (changes requested):
See comments.

rpm/0033-sailfishos-configure-Read-rustc-host-from-environmen.patch
llewelld: No, not sure at all. But I'm also not 100% sure whether this is necessary or not. It'll need investigating.

rpm/xulrunner-qt5.spec
```diff
@@ -356,13 +418,13 @@ echo "ac_add_options --disable-elf-hack" >> "$MOZCONFIG"
 # but as sailfish-browser has privileged EGID, glibc removes it for security reasons. 
 # Set ELF RPATH through LDFLAGS. Needed for plugin-container and libxul.so
 # Additionally we limit the memory usage during linking
- echo 'FIX_LDFLAGS="-Wl,--reduce-memory-overheads -Wl,--no-keep-memory -Wl,-rpath=%{mozappdir}"' >> "${MOZCONFIG}"
+ echo 'FIX_LDFLAGS="-Wl,--strip-debug -Wl,--gc-sections -Wl,--reduce-memory-overheads -Wl,--no-keep-memory -Wl,-rpath=%{mozappdir}"' >> "${MOZCONFIG}"
```
llewelld, line 421: Note to self: this needs arm-guarding to avoid debug symbols being stripped from the x86 build.

rpm/xulrunner-qt5.spec
```diff
@@ -277,10 +304,45 @@ export SB2_RUST_USE_REAL_EXECVP=Yes
...
+echo "export CARGO_BUILD_TARGET=armv7-unknown-linux-gnueabihf" >> "%BUILD_DIR"/rpm-shared.env
+echo "export CARGO_CFG_TARGET_ARCH=arm" >> "%BUILD_DIR"/rpm-shared.env
+
+%build
+
+source "%BUILD_DIR"/rpm-shared.env
+
+#%ifarch %arm
+echo "ac_add_options --host=armv7-unknown-linux-gnueabihf" >> "$MOZCONFIG"
+echo "ac_add_options --target=armv7-unknown-linux-gnueabihf" >> "$MOZCONFIG"
```
llewelld, line 331: These are currently applied on x86 builds as well, so probably aren't having any effect. They can probably be removed.

rpm/xulrunner-qt5.spec
llewelld: I've switched this back to the original now.

rpm/xulrunner-qt5.spec
llewelld: I've added in aarch64 here and also gone through everything so it's now using `%arm32` and `%arm64`.

rpm/xulrunner-qt5.spec
llewelld: Changed to include `%arm64` as well.

rpm/0033-sailfishos-configure-Read-rustc-host-from-environmen.patch
llewelld: I've overlaid a patch to reverse this change, so we can see whether it still builds without it.

rpm/xulrunner-qt5.spec
llewelld: Done.

rpm/xulrunner-qt5.spec
llewelld: Done (refactored).

llewelld:
Note: commits starting with "s" will be squashed.

llewelld:
The build seems to be progression fine, so I've removed the "Draft" flag.

llewelld:
I'll squash the last four commits before merging.

rpm/0033-sailfishos-configure-Read-rustc-host-from-environmen.patch
llewelld: Just in case it's of interest: I also tried dropping this entire patch, but the build then failed on arm, so I brought it back.

xfade (approved):
This should be enough to get it arm building on SDK and OBS (with the sb2-tools changes for OBS)

rainemak (approved):
Went through changes earlier today with @llewelld ! LGTM

llewelld:
Squashed extra commits, rebased on top of latest sailfishos-esr78.

Thaodan (approved):
LGTM

rainemak:
When going forward we should try to fix the tooling so that these kind things would not be needed or at least we would get rid of most of these. For now I think that this is good enough.

okodron (approved):

PR #28, opened 2021-08-26
[sailfishos][embedlite] Use nsWebBrowser to focus active window. Fixes JB#55299 OMP#JOLLA-315

rainemak:
No need to handle window raising / lowering directly via focusmanager.


rainemak:
Rebased

PR #31, opened 2021-08-31
[sailfishos][packaging] Add explicit nss dependencies to the devel package. Fixes JB#55337 OMP#JOLLA-332

adenexter:


adenexter:
The implicit dependency is not being created because the nss libraries are part of the privlibs definitions `# Private/bundled libs the final package should not provide or depend on.` but then the system libraries are used contrary to that.

I did try removing the problem libraries from privlibs but there still wasn't an implicit dependency created.

rpm/xulrunner-qt5.spec
```diff
@@ -234,6 +234,10 @@ Mozilla XUL runner
 Requires: %{name} = %{version}-%{release}
 Conflicts: xulrunner-devel
 Summary: Headers for xulrunner
+# Auto dependency is not picking this up.
+%if %{system_nss}
+Requires:  pkgconfig(nss) >= 3.53.1
```
rainemak, line 239:
```suggestion
Requires: pkgconfig(nss) >= 3.53.1
```
Nit pick, one space would be enough. Doesn't matter that much => your call.

rainemak (approved):

okodron (approved):

PR #40, opened 2021-09-03
[sailfishos][packaging] Remove duplicated SB2 env vars. JB#52891 OMP#JOLLA-92

xfade:
This should have been removed when these statements were moved to a section guarded by arm32 and arm64. It causes warnings of i486 builds.


okodron (approved):

rainemak (approved):
Please rebase before merging.

PR #44, opened 2021-09-13
[sailfishos][gecko] Fix delivery of MozAfterPaint events. JB#55286 OMP#JOLLA-310

adenexter:
Revert a patch to remove an ensure eventual after paint event timer as
no event is delivered in embedlite without it.

The short explanation is that the root widget (an embedlite puppet
widget) of the PresContext where the event is scheduled, is not the same
root widget as in the ClientLayerManager (nsWindow) where the paint is
completed, so the nsView installed as the widget listener on the puppet
widget doesn't have the DidCompositeWindow called via
ClientLayerManager::DidComposite. And even with a redirect established
there's a subsequent failure because the DidComposite is a posted event
and the transaction it refers to is gone when it executes.

### DIFF
--- a/rpm/xulrunner-qt5.spec
+++ b/rpm/xulrunner-qt5.spec
@@ -90,6 +90,7 @@ Patch41:    0041-sailfishos-gecko-Fix-gfxPlatform-AsyncPanZoomEnabled.patch
 Patch42:    0042-sailfishos-gecko-Avoid-incorrect-compiler-optimisati.patch
 Patch43:    0043-sailfishos-gecko-Drop-static-casting-from-BrowserChi.patch
 Patch44:    0044-sailfishos-docshell-Get-ContentFrameMessageManager-v.patch
+Patch45:    0045-Revert-Bug-1445570-Remove-EnsureEventualAfterPaint-t.patch
 #Patch10:    0010-sailfishos-gecko-Remove-PuppetWidget-from-TabChild-i.patch
 #Patch11:    0011-sailfishos-gecko-Make-TabChild-to-work-with-TabChild.patch
 #Patch12:    0012-sailfishos-build-Fix-build-error-with-newer-glibc.patch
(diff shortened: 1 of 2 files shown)


rpm/0001-Revert-Bug-1445570-Remove-EnsureEventualAfterPaint-t.patch
```diff
@@ -0,0 +1,478 @@
+From f745c5ed16d07e607ec355d09e8be1aa3299313b Mon Sep 17 00:00:00 2001
```
rainemak, line 1: Use --zero-commit in addition to none number commit (-N)

rpm/xulrunner-qt5.spec
```diff
@@ -90,6 +90,7 @@ Patch41:    0041-sailfishos-gecko-Fix-gfxPlatform-AsyncPanZoomEnabled.patch
 Patch42:    0042-sailfishos-gecko-Avoid-incorrect-compiler-optimisati.patch
 Patch43:    0043-sailfishos-gecko-Drop-static-casting-from-BrowserChi.patch
 Patch44:    0044-sailfishos-docshell-Get-ContentFrameMessageManager-v.patch
+Patch45:    0001-Revert-Bug-1445570-Remove-EnsureEventualAfterPaint-t.patch
```
rainemak, line 93:
```suggestion
Patch45:    0045-Revert-Bug-1445570-Remove-EnsureEventualAfterPaint-t.patch
```
Rename patch to 0045

rainemak:
I have used "[sailfishos][embedlite]" to ref embedding/embedlite i.e. ones that are not patches. As this is revert and patch this could be "[sailfishos][gecko]".

rainemak (approved):

okodron (approved):

rainemak:
Took a liberty to rebase this one. CC @adenexter

PR #47, opened 2021-09-15
Disable startupCache for all architectures. Fixes JB#55487 OMP-JOLLA-373

llewelld:
Running with a startupCache can cause unexpected behaviour when old code
in the cache is used instead of updated code.

This is intended as a temporary change during development.


rainemak (approved):
Why draft? I think this looks fine.

llewelld:
> Why draft? I think this looks fine.
Ah, okay, I have "undrafted" it :) Thanks.

okodron (approved):

paveltkv (approved):

llewelld:
Rebased.

PR #48, opened 2021-09-15
Avoid stripping debug symbols during final link step on arm. JB#55074 OMP#JOLLA-269

llewelld:
Previously the final link step was failing on arm builds due to
out-of-memory issues (it seemed to run out of address space on the
32-bit SDK).

Something seems to have changed, since it now passes through okay, so
this change makes it so that it no longer strip the debug symbols.


rainemak (approved):
Sounds and looks good. I have successfully build also similarly.

okodron (approved):

paveltkv (approved):

PR #54, opened 2021-09-20
[sailfishos][packaging] Workaround for .git directory in submodule JB#54219 OMP#JOLLA-136

xfade:
Conditional for using a workaround which moves the .git away before build
in order to prevent cargo from updating git modules
This should only be needed for local builds on SDK as tar_git strips out
the .git directory already.

Usage: --with git_workaround

### DIFF
--- a/rpm/xulrunner-qt5.spec
+++ b/rpm/xulrunner-qt5.spec
@@ -38,6 +38,13 @@
 %global __provides_exclude ^(%{privlibs})\\.so
 %global __requires_exclude ^(%{privlibs})\\.so
 
+# Conditional for using a workaround which moves the .git away before build
+# in order to prevent cargo from updating git modules
+# This should only be needed for local builds on SDK as tar_git strips out
+# the .git directory already.
+%bcond_with git_workaround
+
+
 Name:       xulrunner-qt5
 Summary:    XUL runner
 Version:    %{greversion}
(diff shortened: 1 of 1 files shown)


rpm/xulrunner-qt5.spec
```diff
@@ -331,6 +338,12 @@ done
 
 %build
 
+# Move the .git directory out of the way as cargo gets confused and thinks it
+# needs to update our submodule.
+%if %{with git_workaround}
+%__mv %_builddir/.git %_builddir/.git-disabled ||:
```
martyone, line 344: If nothing else is really possible to prevent cargo doing that, then I'd suggest at least

```suggestion
trap '%__mv %_builddir/.git-disabled %_builddir/.git ||:' EXIT
%__mv %_builddir/.git %_builddir/.git-disabled ||:
```

:) and omit the restoration below.

okodron (approved):

rpm/xulrunner-qt5.spec
martyone: Isn't it possible to disable that with `.cargo/config.toml`? https://doc.rust-lang.org/cargo/reference/config.html#netoffline

```
[net]
offline = true
```

rainemak: We already have echo

> "export CARGO_NET_OFFLINE=1" >> "%BUILD_DIR"/rpm-shared.env

Which should be effectively the same.

rpm/xulrunner-qt5.spec
```diff
@@ -441,6 +454,7 @@ echo 'LDFLAGS="$FIX_LDFLAGS"' >> "${MOZCONFIG}"
 echo 'export WRAP_LDFLAGS="$FIX_LDFLAGS"' >> "${MOZCONFIG}"
 echo 'mk_add_options LDFLAGS="$FIX_LDFLAGS"' >> "${MOZCONFIG}"
 
+
```
rainemak, line 457: Extra empty line.

rainemak (approved):

PR #57, opened 2021-09-24
[sailfishos][embedlite] Set BrowsingContext to private mode before attaching it. JB#55492 OMP#JOLLA-377

mkenttala:


embedding/embedlite/embedshared/EmbedLiteViewChild.cpp
```diff
@@ -223,7 +223,7 @@ EmbedLiteViewChild::InitGeckoWindow(const uint32_t parentId, const bool isPrivat
 
   // Create a BrowsingContext for our windowless browser.
   RefPtr<BrowsingContext> browsingContext = BrowsingContext::CreateIndependent(
-              BrowsingContext::Type::Content);
+    isPrivateWindow ? BrowsingContext::Type::Chrome : BrowsingContext::Type::Content);
```
rainemak, line 226: Can you elaborate a bit more? Sounds wrong.

As this seems to be private browsing related crash. Maybe just setting browsing context to private mode is enough:

> browsingContext->SetUsePrivateBrowsing(isPrivateWindow);

mkenttala: docshell/base/BrowsingContext.cpp bool BrowsingContext::CanSetOriginAttributes()

// Attached content BrowsingContexts may have been synced to other processes.
  if (NS_WARN_IF(IsContent())) {
    MOZ_CRASH();
    return false;
  }

That crashes every time if Type::Content is used, so for private mode Type::Chrome must be used. Browser seems to work just fine if the chrome content type is used all the time but I don't know if there are some reasons behind this.

rainemak: I think you're still looking from wrong angle. IMO we should not claim that this very gecko window that we're creating here in this method would be a chrome as it is not chrome window. It is content window and root of the document. Conceptually we do not have chrome window at all.

If you grep nsIDocShellTreeItem::typeChrome, you will start finding differences - as an example does focus handling work and caret moving within an input field. Previously we needed for instance do patch fullscreen api to work but from opposite reason: https://github.com/sailfishos/gecko-dev/blob/master/rpm/0021-sailfishos-gecko-Make-fullscreen-enabling-work-as-us.patch

mkenttala: New approach which does not need that chrome content type, private mode just has to be set before attaching,

rainemak (approved):
Now this looks good.

okodron (approved):

PR #59, opened 2021-09-28
[sailfishos][embedlite] Fix a crash when trying to a acquire a wake lock. Fixes JB#55481 OMP#JOLLA-368

adenexter:


rainemak (approved):
Yeap, clearly missing hal::init => approving.

PR #62, opened 2021-09-28
[sailfishos][egl] Drop swap_buffers_with_damage extension support. Fixes JB#55571 OMP#JOLLA-396

rainemak:
At least on "Adreno (TM) 508" swapping with damage triggered a hang on the
compositor thread i.e. call never returns. Conceptually I think we should
bind eglSwapBuffersWithDamage on hybris side to the eglSwapBuffers or something like that.

Co-authored-by: David Llewellyn-Jones <david.llewellyn-jones@jolla.com>
Signed-off-by: Raine Makelainen <raine.makelainen@jolla.com>

### DIFF
--- a/rpm/0048-sailfishos-egl-Drop-swap_buffers_with_damage-extensi.patch
+++ b/rpm/0048-sailfishos-egl-Drop-swap_buffers_with_damage-extensi.patch
@@ -1,0 +1,41 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Raine Makelainen <raine.makelainen@jolla.com>
+Date: Tue, 28 Sep 2021 09:42:30 +0300
+Subject: [PATCH] [sailfishos][egl] Drop swap_buffers_with_damage extension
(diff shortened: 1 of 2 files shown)


llewelld (approved):
Great! LGTM.

krnlyng:
@rainemak it's probably a libhybris bug.

https://github.com/libhybris/libhybris/blob/master/hybris/egl/egl.c#L412

we should fix it in libhybris - but yes in worst case if it's a driver bug - which i doubt - map eglSwapBuffersWithDamageEXT to eglSwapBuffers.

okodron (approved):

adenexter:
> https://github.com/libhybris/libhybris/blob/master/hybris/egl/egl.c#L412
> 
That raises more questions than it answers, eglSwapBuffers is just a shim around _my_eglSwapBuffersWithDamageEXT. I guess the difference being ws_prepareSwap isn't supplied with any damage rectangles when eglSwapBuffers is called.

adenexter (approved):

rainemak:
Rebased.

PR #61, opened 2021-09-28
Adapt, apply or drop ESR60 patches for ESR78. JB#55228 OMP#JOLLA-298

llewelld:
These changes adapt, apply or drop patches that were applied to ESR60 in the context of ESR78.

PR #68, opened 2021-10-04
[sailfishos][embedlite] Fix child actor services not receiving events. JB#55484 OMP#JOLLA-371

adenexter:
The embedlite message manager needs to be registered with JSActorService. And then the mobile variant of the controls should be favored for Sailfish.

### DIFF
--- a/embedding/embedlite/utils/BrowserChildHelper.cpp
+++ b/embedding/embedlite/utils/BrowserChildHelper.cpp
@@ -19,6 +19,7 @@
 #include "mozilla/dom/MessageManagerBinding.h"
 #include "mozilla/dom/ipc/StructuredCloneData.h"
 #include "mozilla/dom/DocumentInlines.h"
+#include "mozilla/dom/JSActorService.h"
 
 #include "nsNetUtil.h"
 #include "nsIDOMWindowUtils.h"
@@ -260,6 +261,9 @@ BrowserChildHelper::InitBrowserChildHelperMessageManager()
     return false;
   }
   root->SetParentTarget(scope);
+
+  RefPtr<JSActorService> wasvc = JSActorService::GetSingleton();
+  wasvc->RegisterChromeEventTarget(scope);
 
   return true;
 }
--- a/rpm/0050-sailfishos-gecko-Force-use-of-mobile-video-controls..patch
+++ b/rpm/0050-sailfishos-gecko-Force-use-of-mobile-video-controls..patch
@@ -1,0 +1,44 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Andrew den Exter <andrew.den.exter@jolla.com>
+Date: Fri, 1 Oct 2021 06:57:53 +0000
+Subject: [PATCH] [sailfishos][gecko] Force use of mobile video controls.
+ JB#55484 OMP#JOLLA-371
+
+---
+ toolkit/content/widgets/videocontrols.js | 6 +++---
+ 1 file changed, 3 insertions(+), 3 deletions(-)
+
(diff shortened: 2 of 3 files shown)


rpm/0049-sailfishos-gecko-Enable-the-mobile-video-control-con.patch
```diff
@@ -0,0 +1,47 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Andrew den Exter <andrew.den.exter@jolla.com>
+Date: Fri, 1 Oct 2021 06:57:53 +0000
+Subject: [PATCH] [sailfishos][gecko] Enable the mobile video control config
+ for Sailfish.
```
rainemak, line 5:
```suggestion
 for Sailfish. JB#55484 OMP#JOLLA-371
```
Good to  have bug references here as well.

rpm/0049-sailfishos-gecko-Enable-the-mobile-video-control-con.patch
```diff
@@ -0,0 +1,47 @@
...
+index 0b41769f680c..03e98a3c01d0 100644
+--- a/toolkit/content/widgets/videocontrols.js
++++ b/toolkit/content/widgets/videocontrols.js
+@@ -19,7 +19,8 @@ this.VideoControlsWidget = class {
+     this.document = this.element.ownerDocument;
+     this.window = this.document.defaultView;
+ 
+-    this.isMobile = this.window.navigator.appVersion.includes("Android");
++    this.isMobile = this.window.navigator.appVersion.includes("Android") ||
++                    this.window.navigator.appVersion.includes("Sailfish");
```
rainemak, line 21: This requires that start using user-agent building patch which is fine. It should ok to land this before https://github.com/sailfishos/gecko-dev/pull/61/ . Alternatively you could pick that user-agent building to this PR which would be logically good option as well. Please sync with @llewelld how to handle.

rainemak:
Very nice fix!

embedding/embedlite/utils/BrowserChildHelper.cpp
```diff
@@ -261,6 +262,9 @@ BrowserChildHelper::InitBrowserChildHelperMessageManager()
   }
   root->SetParentTarget(scope);
 
+  RefPtr<JSActorService> wasvc = JSActorService::GetSingleton();
+  wasvc->RegisterChromeEventTarget(scope);
```
rainemak, line 266: Did you check how it goes in nsFrameLoader::EnsureMessageManager? Does that InProcessBrowserChildMessageManager::Create get called ?

adenexter: That was the eventual source of the fix.

rainemak (approved):

rainemak:
Please rebased

okodron (approved):

rpm/0049-sailfishos-gecko-Enable-the-mobile-video-control-con.patch
```diff
@@ -0,0 +1,47 @@
...
```
rainemak, line 21: We could consider this to be like

> this.isMobile = true;

okodron: That's better.

adenexter: Done

rpm/0049-sailfishos-gecko-Enable-the-mobile-video-control-con.patch
adenexter: I've changed it (back) to always be true as suggested.

PR #69, opened 2021-10-05
[sailfishos][audio] Bring back "pulse" cubeb backend. Fixes JB#55765 OMP#JOLLA-423

rainemak:
It seems that "pulse-rust" is not working for us for some reason. Didn't
start digging that too deeply. Worth noting that there's also "kai"
backend which might be worth checking when going forward.

Let's do not merge this one before https://github.com/sailfishos/gecko-dev/pull/61 is merged. After https://github.com/sailfishos/gecko-dev/pull/61 is merged, I'll rebase this and adjust patch numbers.

### DIFF
--- a/rpm/0074-Revert-Bug-1611658-Unship-libcubeb-s-C-PulseAudio-ba.patch
+++ b/rpm/0074-Revert-Bug-1611658-Unship-libcubeb-s-C-PulseAudio-ba.patch
@@ -1,0 +1,1655 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Raine Makelainen <raine.makelainen@jolla.com>
+Date: Tue, 5 Oct 2021 10:58:27 +0300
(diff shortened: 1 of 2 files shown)


rainemak:
Rebased

adenexter:
> This is to make volume controlling to work. Everything that had less than 1.0 volume level are going to be muted.
> 
> For example here: https://browser.sailfishos.org/tests/testvideo.html
> 
> Start "Big Buck Bunny" playback and drag that slider below the video element right below "Initial volume". 

I understand. I couldn't reproduce this earlier this week with the video control changes prior to them being merged so I'm trying to figure out what's different, one possibility being the submodule was pinned to an earlier commit before the pulse backed was dropped. It's not the additional patches being applied because playback now playback is crashing.

rainemak:
> One possibility being the submodule was pinned to an earlier commit before the pulse backed was dropped. It's not the additional patches being applied because playback now playback is crashing.

Sorry, now I understood. Also with sha1 f8fd279cca66fc80793fd55ec520c79eab571f80 I updated last time submodule sha1.

PR #70, opened 2021-10-07
[sailfishos][gecko] Use only embedlite clipboard. JB#55607 OMP#JOLLA-399

mkenttala:


mkenttala:
Now patch 0002 modified to not to add Qt clipboard.

PR #83, opened 2021-10-20
Reformat patch to allow git am to work

llewelld:
Minor errors in the patch prevented it from being applied in this way.
This reformatting fixes it so that

git am rpm/*.patch

will now work correctly.


rainemak:
Please rebase. Let's take this also along with the next tag.

llewelld:
Rebased.

PR #91, opened 2021-10-26
[sailfishos][gecko] Allow LoginManagerPrompter to find its window. JB#55760 OMP#JOLLA-418

llewelld:
This patch blocks loading of gecko's LoginManagerAuthPrompter.jsm so that the embedlite-components version can be used instead.

It also patches the `nsILoginManagerPrompter` interface to allow a reference to the window to be passed through, to allow the embedlite component to understand its context.

### DIFF
--- a/rpm/0083-sailfishos-gecko-Allow-LoginManagerPrompter-to-find-.patch
+++ b/rpm/0083-sailfishos-gecko-Allow-LoginManagerPrompter-to-find-.patch
@@ -1,0 +1,120 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: David Llewellyn-Jones <david.llewellyn-jones@jolla.com>
+Date: Thu, 21 Oct 2021 12:35:53 +0000
+Subject: [PATCH] [sailfishos][gecko] Allow LoginManagerPrompter to find its
+ window. JB#55760, OMP#JOLLA-418
+
+This patch blocks loading of gecko's LoginManagerAuthPrompter.jsm so
+that the embedlite-components version can be used instead.
+
+It also patches the nsILoginManagerPrompter interface to allow a
+reference to the window to be passed through, to allow the embedlite
+component to understand its context.
+
+Finally it patches ParentChannelListener to pass the correct window
+object through to the nsILoginManagerAuthPrompter component.
+---
+ netwerk/protocol/http/ParentChannelListener.cpp      |  2 +-
+ .../components/passwordmgr/LoginManagerParent.jsm    |  4 ++--
+ toolkit/components/passwordmgr/components.conf       | 12 ------------
+ .../passwordmgr/nsILoginManagerPrompter.idl          |  6 +++---
+ 4 files changed, 6 insertions(+), 18 deletions(-)
+
+diff --git a/netwerk/protocol/http/ParentChannelListener.cpp b/netwerk/protocol/http/ParentChannelListener.cpp
+index 39f73f998e74..80a25d59628c 100644
+--- a/netwerk/protocol/http/ParentChannelListener.cpp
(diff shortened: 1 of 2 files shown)


llewelld:
For testing, this should be used alongside https://github.com/sailfishos/embedlite-components/pull/64

rpm/0082-sailfishos-gecko-Allow-LoginManagerPrompter-to-find-.patch
```diff
@@ -0,0 +1,120 @@
...
+diff --git a/netwerk/protocol/http/ParentChannelListener.cpp b/netwerk/protocol/http/ParentChannelListener.cpp
+index 39f73f998e74..80a25d59628c 100644
+--- a/netwerk/protocol/http/ParentChannelListener.cpp
++++ b/netwerk/protocol/http/ParentChannelListener.cpp
+@@ -401,7 +401,7 @@ ParentChannelListener::GetAuthPrompt(uint32_t aPromptReason, const nsIID& iid,
+ 
+   nsCOMPtr<nsPIDOMWindowOuter> window;
+   RefPtr<dom::Element> frame = mBrowsingContext->Top()->GetEmbedderElement();
+-  if (frame) window = frame->OwnerDoc()->GetWindow();
++  window = frame ? frame->OwnerDoc()->GetWindow(): mBrowsingContext->GetDOMWindow();
```
rainemak, line 32:
```suggestion
+  window = frame ? frame->OwnerDoc()->GetWindow() : mBrowsingContext->GetDOMWindow();
```

rpm/0082-sailfishos-gecko-Allow-LoginManagerPrompter-to-find-.patch
```diff
@@ -0,0 +1,120 @@
...
+diff --git a/toolkit/components/passwordmgr/LoginManagerParent.jsm b/toolkit/components/passwordmgr/LoginManagerParent.jsm
+index ab4b36b6161a..b143758f878c 100644
+--- a/toolkit/components/passwordmgr/LoginManagerParent.jsm
++++ b/toolkit/components/passwordmgr/LoginManagerParent.jsm
+@@ -804,7 +804,7 @@ class LoginManagerParent extends JSWindowActorParent {
+       }
+     }
+ 
+-    let promptBrowser = LoginHelper.getBrowserForPrompt(browser);
++    let promptBrowser = browsingContext.window;
```
rainemak, line 45: A question, would it be simpler or possible to patch LoginHelper.jsm instead?

rainemak:
Added two small comments / notes. Please give a look to those. Also please rebase this one, patch number should be now 83.

rpm/0082-sailfishos-gecko-Allow-LoginManagerPrompter-to-find-.patch
llewelld: I thought about it too, but in practice it would result in even greater changes here, since at this point in the code `browser` is null, since we're missing the `embedderElement` returned by `getRootBrowser()`, so we'd need to pass a different value to `getBrowserForPrompt(browser)`.

If at some point in the future `embedderElement` is set to a value, then I think your suggestion would definitely make sense.

llewelld:
Rebased.

llewelld:
Patch updated to 83.

rpm/0082-sailfishos-gecko-Allow-LoginManagerPrompter-to-find-.patch
rainemak: Ok, sounds reasonable.

rainemak (approved):
Approving already. Please rebase before merging as there might other PRs going in as well.

okodron (approved):

llewelld:
Rebased; removed comma from commit title.

PR #96, opened 2021-10-29
[sailfishos][embedlite] Remove unused variable. Contributes to JB#55799

chriadam:
Previously, the variable existed because a template function could
not accept `nullptr` as a parameter; this was fixed upstream, and
the code was updated to pass `nullptr` directly, but the variable
was never removed, resulting in compiler warning.


chriadam:
Rebased.

PR #97, opened 2021-11-01
Fix Security info (secure/unsecure) to appear to the toolbar

rainemak:
Contains 3 commits:

1. First one makes it possible to use nsSecureBrowserUI and update it
when location is changed. Follow up commit will create nsSecureBrowserUI.
2. Fix crash in EmbedLiteSecurity
3. Create Secure Browser UI when gecko window created

### DIFF
--- a/embedding/embedlite/embedshared/EmbedLiteViewChild.cpp
+++ b/embedding/embedlite/embedshared/EmbedLiteViewChild.cpp
@@ -37,6 +37,7 @@
 #include "nsIWidgetListener.h"
 #include "mozilla/layers/APZEventState.h"
 #include "APZCCallbackHelper.h"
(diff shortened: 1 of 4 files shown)


okodron:
Renamed `rpm/0077-sailfishos-docshell-Align-fix-parentprocess-checks.-.patch` should be also changed in spec.

rainemak:
> Renamed `rpm/0077-sailfishos-docshell-Align-fix-parentprocess-checks.-.patch` should be also changed in spec.

Done.

okodron (approved):

llewelld:
This all LGTM, and also worked nicely when tested too.

llewelld (approved):

rainemak:
Rebased

PR #98, opened 2021-11-01
[sailfishos][embedlite] Set ScreenManager's screen properties. JB#56165 OMP#JOLLA-502

mkenttala:


embedding/embedlite/embedshared/EmbedLiteWindowChild.cpp
```diff
@@ -127,17 +126,10 @@ mozilla::ipc::IPCResult EmbedLiteWindowChild::RecvSetContentOrientation(const ui
...
-      do_GetService("@mozilla.org/gfx/screenmanager;1", &rv);
-  NS_ENSURE_TRUE(screenMgr, IPC_OK());
-
-  nsIntRect rect;
   int32_t colorDepth, pixelDepth;
   nsCOMPtr<nsIScreen> screen;
 
-  screenMgr->GetPrimaryScreen(getter_AddRefs(screen));
-  screen->GetRect(&rect.x, &rect.y, &rect.width, &rect.height);
+  ScreenManager::GetSingleton().GetPrimaryScreen(getter_AddRefs(screen));
```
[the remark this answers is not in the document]
rainemak: Only UIKit still has specialized implementation of nsIScreenManager. I'd expect that rest would go with ScreenManager of ScreenManager.h. That said we could check if we can drop nsScreenManagerQt => separate issue.

embedding/embedlite/embedshared/EmbedLiteWindowChild.cpp
```diff
@@ -199,9 +196,21 @@ void EmbedLiteWindowChild::CreateWidget()
...
+
   mInitialized = true;
   Unused << SendInitialized();
 }
 
+void EmbedLiteWindowChild::RefreshScreen(int width, int height)
+{
+  LayoutDeviceIntRect rect(0, 0, width, height);
+  AutoTArray<RefPtr<Screen>, 1> screenList;
+  RefPtr<Screen> screen = new Screen(rect, rect, 0, 0, DesktopToLayoutDeviceScale(), CSSToLayoutDeviceScale(), 96);
```
rainemak, line 210: These initial values feel wrong. I think we should pass down to the EmbedLiteWindowChild (from QtMozEmbed via EmbedLiteWindowParent) at least dpi and screen color depth. ScreenHelperAndroid::AddScreen uses 24bits for pixel and color depth and has fixed 160.0 for dpi.

rainemak: Surely we can start with this one create more tasks on top of this.

rainemak (commented):
Added some notes. Please give a look.

embedding/embedlite/embedshared/EmbedLiteWindowChild.cpp
```diff
@@ -199,9 +196,42 @@ void EmbedLiteWindowChild::CreateWidget()
...
+
+void EmbedLiteWindowChild::SetDepth(int depth)
+{
+  if (depth != mDepth) {
+    mDepth = depth;
+    RefreshScreen();
+  }
+}
+
+void EmbedLiteWindowChild::SetDensity(float density)
```
rainemak, line 228: I'd rather have

```
SetScreenProperties(int depth, float density, float dpi)
```
=> That would imply one send from parent to child + only one RefreshScreen()

rainemak: I think this is good enough now. One improvement could be that we'd deliver window properties directly to the window without using view in between. Thanks @mkenttala for adjusting. As said, good enough.

mkenttala:
Combined the screen properties but didn't dare to mess with the current SetDPI as it seems to go through several layers and even contains patching, not sure if there's something vital there.

embedding/embedlite/embedshared/EmbedLiteWindowChild.cpp
```diff
@@ -199,9 +196,44 @@ void EmbedLiteWindowChild::CreateWidget()
...
+void EmbedLiteWindowChild::RefreshScreen()
+{
+  LayoutDeviceIntRect rect;
+  if (mRotation == ROTATION_0 || mRotation == ROTATION_180)
+    rect = mBounds;
+  else
+    rect = LayoutDeviceIntRect(0, 0, mBounds.Height(), mBounds.Width());
+
+  AutoTArray<RefPtr<Screen>, 1> screenList;
+  RefPtr<Screen> screen = new Screen(rect, rect, mDepth, mDepth, DesktopToLayoutDeviceScale(mDensity), CSSToLayoutDeviceScale(1.0f), GetWidget()->GetDPI());
```
rainemak, line 215: It looks to me that this GetDPI() is not the right one as this GetWidget() is nsWindow where as EmbedLitePuppetWidget::GetDPI reads back the dpi from the EmbedLiteViewChild.

I'd say that simpler to pass down dpi to here within SetScreenProperties

rainemak (approved):

okodron (approved):

PR #99, opened 2021-11-05
[sailfishos][gecko] Fix download starting when multiple tabs are closed. JB#56047 OMP#JOLLA-476

adenexter:
When the active tab is closed another tab will be made active and start
to load, if that tab is closed while loading the parent puppet widget can be
destroyed before the content view is created causing a failure in the
content view construction the error from which gets interpreted as the
unsupported document type and the browser falls back to downloading.

There is a failure case for when this happens and the DocShell is in
destroyed state which cancels the loading the document, so we can do the
same if the puppet widget has been destroyed.

### DIFF
--- a/embedding/embedlite/embedshared/EmbedLiteViewChild.cpp
+++ b/embedding/embedlite/embedshared/EmbedLiteViewChild.cpp
@@ -183,6 +183,9 @@ mozilla::ipc::IPCResult EmbedLiteViewChild::RecvDestroy()
   }
 
   EmbedLiteAppService::AppService()->UnregisterView(mId);
+  if (mWebBrowser) {
+    mWebBrowser->Destroy();
+  }
   if (mHelper)
     mHelper->Unload();
   if (mChrome)


rpm/0084-sailfishos-gecko-Fix-download-starting-when-multiple.patch
```diff
@@ -0,0 +1,48 @@
...
+diff --git a/docshell/base/nsDocShell.cpp b/docshell/base/nsDocShell.cpp
+index ae8b77a80cdc..912ca0b67a5a 100644
+--- a/docshell/base/nsDocShell.cpp
++++ b/docshell/base/nsDocShell.cpp
+@@ -7659,7 +7659,8 @@ nsresult nsDocShell::CreateContentViewer(const nsACString& aContentType,
+     }
+   }
+ 
+-  NS_ENSURE_SUCCESS(Embed(viewer), NS_ERROR_FAILURE);
++  rv = Embed(viewer)
```
rainemak, line 29:
```suggestion
+  rv = Embed(viewer);
```

rpm/0084-sailfishos-gecko-Fix-download-starting-when-multiple.patch
```diff
@@ -0,0 +1,48 @@
...
++  NS_ENSURE_SUCCESS(rv, rv);
```
rainemak, line 30: This looks like a void to me.

adenexter: I don't know what you mean.

The change itself is in the event of an error returning the error value from Embed() instead of NS_ERROR_FAILURE allowing the NS_ERROR_DOCSHELL_DYING from SetupNewViewer() to reach nsDSURIContentListener::DoContent() where the content download will be cancelled and reported as being successfully handled instead of returning a failure which then results in a fallback download handler for the content.

okodron (approved):

rpm/0084-sailfishos-gecko-Fix-download-starting-when-multiple.patch
```diff
@@ -0,0 +1,48 @@
...
+ 
+   if (!mBrowsingContext->GetHasLoadedNonInitialDocument()) {
+     mBrowsingContext->SetHasLoadedNonInitialDocument(true);
+@@ -7853,7 +7854,10 @@ nsresult nsDocShell::SetupNewViewer(nsIContentViewer* aNewViewer,
+ 
+   mContentViewer->SetNavigationTiming(mTiming);
+ 
+-  if (NS_FAILED(mContentViewer->Init(widget, bounds, aWindowActor))) {
++  if (widget->Destroyed()) {
```
rainemak, line 39:
```suggestion
+  if (widget && widget->Destroyed()) {
```
That's bare minimum needed change. I'd try to check if could destroy and clear docshell and friends in the EmbedLiteViewChild::RecvDestroy

adenexter: Calling mWebBrowser->Destroy(); in RecvDestroy worked for me closing a dozen tabs in a row. And I put some debug in nsDocShell::Destroy() and saw that it was being called. I also tried just clearing the mWebBrowser pointer but I didn't see my debug statements and then I had a crash.

rainemak: Calling mWebBrowser->Destroy(); in RecvDestroy of EmbedLiteViewChild a assume. If so, that would be better as then we can avoid one additional patch.

rainemak (approved):
This is cool and makes sense! No need to patch docshell.

okodron (approved):

okodron:
Even better.

PR #103, opened 2021-11-10
[sailfishos][gecko] Fix no audio during WebRTC call. JB#56302

d-grigorev:
During call setup, the audio buffer latency may become high because
the engine does a lot of work. Due to a bug in libcubeb. it may start
dropping audio data instead of writing it to the pulseaudio sink input.

### DIFF
--- a/rpm/0072-sailfishos-gecko-Fix-audio-underruns-for-fullduplex-.patch
+++ b/rpm/0072-sailfishos-gecko-Fix-audio-underruns-for-fullduplex-.patch
@@ -4,12 +4,12 @@ Date: Fri, 3 Sep 2021 13:07:18 +0300
 Subject: [PATCH] [sailfishos][gecko] Fix audio underruns for fullduplex mode
 
 ---
- media/libcubeb/src/cubeb_pulse.c              | 421 +++++++++++++++++-
+ media/libcubeb/src/cubeb_pulse.c              | 437 +++++++++++++++++-
  .../src/peerconnection/PeerConnectionImpl.cpp |  12 +-
- 2 files changed, 407 insertions(+), 26 deletions(-)
+ 2 files changed, 423 insertions(+), 26 deletions(-)
 
 diff --git a/media/libcubeb/src/cubeb_pulse.c b/media/libcubeb/src/cubeb_pulse.c
-index 1cac5eadd45b..13a205e5383d 100644
+index 1cac5eadd45b..89afce055a4f 100644
 --- a/media/libcubeb/src/cubeb_pulse.c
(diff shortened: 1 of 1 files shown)


rpm/0072-sailfishos-gecko-Fix-audio-underruns-for-fullduplex-.patch
```diff
@@ -250,15 +256,19 @@ index 1cac5eadd45b..13a205e5383d 100644
...
++      // otherwise the latency calculator will get stuck.
 +      r = WRAP(pa_stream_get_latency)(stm->output_stream, &latency, NULL);
-+      if (r == 0 && latency > 300 * PA_USEC_PER_MSEC) {
-+        cubeb_fifo_read_done(&fdx->playback, playback_len);
-+      } else {
-+        // Write playback data.
-+        cubeb_fullduplex_play(stm);
++      LOGV("Latency: %lld", latency);
++      if (r == 0 && latency > 300 * PA_USEC_PER_MSEC && got > 1) {
++        size_t drop_len = MIN(playback_len / 2, cubeb_fifo_size(&fdx->playback));
```
llewelld, line 265: I can't claim to understand this code, but is it safe for `drop_len` to be something other than a multiple of `fdx->out_frame_size` here?

llewelld (commented):
The code changes all look sane to me, I just have one small comment. Is there any reasonable way to test this?

rpm/0072-sailfishos-gecko-Fix-audio-underruns-for-fullduplex-.patch
d-grigorev: Thank you for pointing me on that. I fixed this in the new version of the commit. Also, just in case, I added a code to handle the error value from the callback.

okodron (approved):
LGTM from the code point of view.

d-grigorev:
> The code changes all look sane to me, I just have one small comment. Is there any reasonable way to test this?

The bug may not appear on a high-perfomance device. I can't think of any ways to test it, except making real video calls on and grep'ing logs of a debug build with MOZ_LOG="cubeb:5" and looking if the code branch works.

rpm/0072-sailfishos-gecko-Fix-audio-underruns-for-fullduplex-.patch
llewelld: Great, thank you!

llewelld (approved):
I wasn't able to test this in practice, but the code changes all look very sensible, and I didn't hit any issues building, so this LGTM.

PR #104, opened 2021-11-10
Drop embedlite specific History listener

rainemak:


okodron (approved):
LGTM

rainemak:
Rebased

rainemak:
Rebased

llewelld (approved):
This is kind-of brutal, but the changes are clear. The only practical difference I could see is that "em:history" will no longer be broadcast, but I also couldn't see anything listening for it, so that's fine. If `EmbedHistoryListener` is no longer needed, then that's really nice.

Building and in brief testing, I didn't experience any issues.

PR #105, opened 2021-11-11
[sailfishos][gecko-dev] Enable the pointer event API. JB#56149 OMP#JOLLA-499

adenexter:
It was disabled because the APZ page zoom gets priority access to pinch
gestures which meant custom zooming using the pointer events didn't work
on some sites, but a fallback to legacy touch APIs did work if the
pointer API wasn't available did. Other sites have the inverse problem
where their fallback code paths have rotted and they don't work without
the pointer API.

The page zoom getting priority issue can be mitigated by disabling it if
the page sets both the meta viewport minimum and maximum scales to 1.

### DIFF
--- a/rpm/0084-sailfishos-gecko-dev-Disallow-page-zooming-if-the-me.patch
+++ b/rpm/0084-sailfishos-gecko-dev-Disallow-page-zooming-if-the-me.patch
@@ -1,0 +1,46 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Andrew den Exter <andrew.den.exter@jolla.com>
+Date: Thu, 11 Nov 2021 06:51:59 +0000
+Subject: [PATCH] [sailfishos][gecko-dev] Disallow page zooming if the meta
+ viewport scale is fixed. JB#56149 OMP#JOLLA-49
+
+---
+ layout/base/nsLayoutUtils.cpp | 16 +++++++++++++++-
+ 1 file changed, 15 insertions(+), 1 deletion(-)
+
+diff --git a/layout/base/nsLayoutUtils.cpp b/layout/base/nsLayoutUtils.cpp
+index 2cd2903a8fc8..85e290c0d90f 100644
+--- a/layout/base/nsLayoutUtils.cpp
++++ b/layout/base/nsLayoutUtils.cpp
+@@ -694,13 +694,27 @@ bool nsLayoutUtils::AsyncPanZoomEnabled(nsIFrame* aFrame) {
+   return widget->AsyncPanZoomEnabled();
+ }
+ 
++
++static bool DocumentHasFixedUnityZoom(const mozilla::dom::Document* aDocument)
++{
++  if (!aDocument) {
++    return false;
++  }
++  nsresult rv;
++  ViewportMetaData metaData = aDocument->GetViewportMetaData();
++  return (!metaData.mMinimumScale.IsEmpty() && metaData.mMinimumScale == metaData.mMaximumScale) ||
++          metaData.mUserScalable.EqualsLiteral("0") ||
++          metaData.mUserScalable.EqualsLiteral("no") ||
++          metaData.mUserScalable.EqualsLiteral("false");
++}
++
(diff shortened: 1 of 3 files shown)


rpm/0084-sailfishos-gecko-dev-Disallow-page-zooming-if-the-me.patch
```diff
@@ -0,0 +1,43 @@
...
+ 
++
++static bool DocumentHasFixedUnityZoom(const mozilla::dom::Document* aDocument)
++{
++  if (!aDocument) {
++    return false;
++  }
++  nsresult rv;
++  ViewportMetaData metaData = aDocument->GetViewportMetaData();
++  return metaData.mMinimumScale.ToDouble(&rv) == 1.0 && metaData.mMaximumScale.ToDouble(&rv) == 1.0;
```
rainemak, line 27: It's enough that min and max are equal to fix zooming. Further, following conditions should disable zooming as well:

```
metaData.mUserScalable.EqualsLiteral("0") or
metaData.mUserScalable.EqualsLiteral("no") or
metaData.mUserScalable.EqualsLiteral("false")
```

Further, there places like (e.g in ZoomContraintsClient.cpp):
```
  constraints.mAllowZoom = aViewportInfo.IsZoomAllowed() &&
                           nsLayoutUtils::AllowZoomingForDocument(aDocument);
```
Hence, thinking that maybe worth adjusting Document::GetViewportInfo as well so that effectiveZoomFlag ends up being nsViewportInfo::ZoomFlag::DisallowZoom also when min & max are equal (Document.cpp ~ 9597).

rainemak:
Also split embedding change and patch into separate commits.

rpm/0084-sailfishos-gecko-dev-Disallow-page-zooming-if-the-me.patch
adenexter: Document::GetViewportInfo returns a ViewportMetaData. nsViewport is returned by Document::GetViewportInfo which requires display dimensions as arguments.

rpm/0084-sailfishos-gecko-dev-Disallow-page-zooming-if-the-me.patch
```diff
@@ -0,0 +1,46 @@
...
+diff --git a/layout/base/nsLayoutUtils.cpp b/layout/base/nsLayoutUtils.cpp
+index 2cd2903a8fc8..85e290c0d90f 100644
+--- a/layout/base/nsLayoutUtils.cpp
++++ b/layout/base/nsLayoutUtils.cpp
+@@ -694,13 +694,27 @@ bool nsLayoutUtils::AsyncPanZoomEnabled(nsIFrame* aFrame) {
+   return widget->AsyncPanZoomEnabled();
+ }
+ 
++
++static bool DocumentHasFixedUnityZoom(const mozilla::dom::Document* aDocument)
```
rainemak, line 20:
```suggestion
+static bool DocumentHasFixedUnityZoom(const mozilla::dom::Document *aDocument)
```

rainemak (approved):
This should work now. Approving.

rpm/0084-sailfishos-gecko-dev-Disallow-page-zooming-if-the-me.patch
rainemak: You missed my point. I meant that effectiveZoomFlag inside Docoment::GetViewportInfo currently does not end up being nsViewportInfo::ZoomFlag::DisallowZoom (Document.cpp:9635) when minimum and maximum are the same (mAllowZoom = false => Document.cpp:9601).

I think this is just in correct behaviour.

okodron (approved):

rainemak:
Just realized that there are two commits, which is good, but content of these commits ain't good. Other should have embedding preferences and other one contain proposed patch.

rainemak:
> Just realized that there are two commits, which is good, but content of these commits ain't good. Other should have embedding preferences and other one contain proposed patch.

I did small adjustment to the commits by soft resetting it and of course I kept the author as it was. Also rebased the change.

PR #117, opened 2021-11-18
[sailfishos][gecko] Fix audio underruns for the fullduplex mode. JB#56317

d-grigorev:
Move audio rendering from the pulseaudio callback to a separate thread.


adenexter:
Looks good, the diff to the esr60 patch only has a couple of differences which appear to be correct fixes.

adenexter (approved):

okodron (commented):
LGTM as well.

PR #121, opened 2021-11-26
Fix mesa egl display and buffer initialisation. Fixes JB#56215 OMP#JOLLA-512

llewelld:
Ensure the same display is used for all initialisations to avoid creating multiple wayland connections.

Fallback to a wayland window surface in case pixel buffers aren't supported. This is needed on the emulator.

This extends Adam Pigg's (@piggz) earlier patch which addressed a similar issue for the browser on non-hybris devices (including the emulator).


rpm/0038-sailfishos-embedlite-egl-Fix-mesa-egl-display-and-bu.patch
```diff
@@ -0,0 +1,488 @@
...
++    NS_WARNING("Error opening libwayland-client.so.0");
++    return false;
++  }
++  _wl_surface_interface = (const struct wl_interface *)dlsym(handle, "wl_surface_interface");
++
++  *(void **)(&_wl_proxy_marshal_constructor) = dlsym(handle, "wl_proxy_marshal_constructor");
++  *(void **)(&_wl_proxy_marshal) = dlsym(handle, "wl_proxy_marshal");
++  *(void **)(&_wl_proxy_destroy) = dlsym(handle, "wl_proxy_destroy");
++
++  dlclose(handle);
```
krnlyng, line 141: It's probably not too big of an issue, but if libwayland-egl.so.1 gets unloaded by a final dlclose (potentially somewhere else) and one of the wayland functions is then accessed this will crash. I think it might be better to move the dlclose after we are sure that the functions won't be accessed anymore. However i think it's probably unlikely that this will cause any issues so i'm fine with it either way.

adenexter (approved):

rpm/0038-sailfishos-embedlite-egl-Fix-mesa-egl-display-and-bu.patch
adenexter: It's going to fail immediately because nothing else loads the library, or never because the wayland QPA plugin will link against it and that won't get unloaded ever.

d-grigorev (approved):

krnlyng (approved):

rpm/0038-sailfishos-embedlite-egl-Fix-mesa-egl-display-and-bu.patch
llewelld: I pushed some changes to avoid the former happening (which may never have happened, but better safe than sorry?). The new changes are in a separate commit which I'll squash before merging.

Now, if the functions get dynamically loaded, they don't get unloaded again until all of the surfaces have been released, which in practice only happens during shutdown of the app.

llewelld:
The pushed changes are to avoid the dynamic wayland libraries getting unloaded during execution. The changes are in a separate commit, which I'll squash before merging.

As before, these changes should only affect devices which don't support pbuffers, and even then only for WebGL on the browser.

llewelld:
Squashed and rebased.

PR #123, opened 2021-12-03
Fix egl patch to prevent crashing. Fixes JB#56606 OMP#JOLLA-565

llewelld:
A static function in the code had return value `bool`, but no return statement. The compiler then optimised this undefined behaviour into a crash (calling an otherwise guarded `dlopen` with an invalid parameter).

This change removes the crash by setting the return as void. Now the `dlopen` is only called when it's supposed to be.

Thanks to @Tomin1 and especially @spiiroin for noticing the incorrect code.

This change also fixes up an incorrectly named variable.


llewelld:
Note the crucial functional change is line 156 (the rest is just renaming):
```c++
- static bool UnloadWaylandFunctions() {
+ static void UnloadWaylandFunctions() {
```

Tomin1 (approved):

llewelld:
Fixed a typo in the commit title.

rainemak (approved):

paveltkv (approved):

okodron (approved):

llewelld:
Many thanks for the swift turnaround on this all.

PR #127, opened 2021-12-14
[sailfishos][embedlite] Fix user agent notify. Fixes JB#56714 OMP#JOLLA-605

mkenttala:
If browser.tabs.documentchannel is enabled first request coming to WebBrowserChrome is prefixed with "documentchannel:" which causes it to fail to notify user-agent changes.

### DIFF
--- a/embedding/embedlite/utils/WebBrowserChrome.cpp
+++ b/embedding/embedlite/utils/WebBrowserChrome.cpp
@@ -24,7 +24,6 @@
 #include "nsIBaseWindow.h"
 #include "nsIMultiPartChannel.h"
 #include "nsIHttpProtocolHandler.h"
-#include "nsIObserver.h"
 #include "mozilla/dom/ScriptSettings.h" // for AutoNoJSAPI
 #include "mozilla/dom/EventTarget.h"
 #include "BrowserChildHelper.h"
@@ -77,7 +76,6 @@ WebBrowserChrome::WebBrowserChrome(nsIEmbedBrowserChromeListener* aListener)
   , mFirstPaint(false)
   , mScrollOffset(0,0)
   , mListener(aListener)
-  , mRequest(nullptr)
 {
   LOGT();
 }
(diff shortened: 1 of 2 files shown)


mkenttala:
From documentchannel.h:
/**
 * DocumentChannel is a protocol agnostic placeholder nsIChannel implementation
 * that we use so that nsDocShell knows about a connecting load. It transfers
 * all data into a DocumentLoadListener (running in the parent process), which
 * will create the real channel for the connection, and decide which process to
 * load the resulting document in. If the document is to be loaded in the
 * current process, then we'll synthesize a redirect replacing this placeholder
 * channel with the real one, otherwise the originating docshell will be removed
 * during the process switch.
 */

So my understanding is that documentchannel sandboxes requests, it's on by default also in android firefox 78. The easy fix for the bug is to disable this feature but if we wan't to keep documentchannel then WebBrowserChrome needs some work to recognize these documentchannel requests.

mkenttala:
This new simplified approach works both with and without documentchannel.

embedding/embedlite/utils/WebBrowserChrome.cpp
```diff
@@ -256,12 +253,16 @@ WebBrowserChrome::OnStateChange(nsIWebProgress* progress, nsIRequest* request,
...
   if (progressStateFlags & nsIWebProgressListener::STATE_START && progressStateFlags & nsIWebProgressListener::STATE_IS_DOCUMENT) {
-    Unused << AddUserAgentObserver(request);
     mListener->OnLoadStarted(mLastLocation.get());
   }
   if (progressStateFlags & nsIWebProgressListener::STATE_STOP && progressStateFlags & nsIWebProgressListener::STATE_IS_DOCUMENT) {
-    Unused << RemoveUserAgentObserver(request);
     mListener->OnLoadFinished();
+    nsAutoString httpUserAgent;
+    nsresult rv = GetHttpUserAgent(request, httpUserAgent);
+    if (NS_SUCCEEDED(rv) && !httpUserAgent.IsEmpty()) {
```
rainemak, line 262: What if the user-agent override is really empty? Pondering do we need that is not empty guard.

mkenttala: During my testing msn.com sometimes loaded about:blank too which had empty ua, added this guard to prevent overwrite.

rainemak: Yeap, there's about:blank when new tab request start. So, that's expected. Thanks for clarifying => let's leave it there.

rainemak:
Please rebase.

rainemak (approved):

okodron (approved):

PR #128, opened 2021-12-14
Ensure clipboard service can be accessed. Fixes JB#56647 OMP#JOLLA-578

llewelld:
The CID used to access the clipboard service in nsCopySupport fails (it doesn't match the CID of nsIClipboard). This prevents the embedlite nsIClipboard service from being accessed.

Using the contract ID instead of the CID allows it to be accessed correctly.

### DIFF
--- a/embedding/embedlite/components/nsClipboard.h
+++ b/embedding/embedlite/components/nsClipboard.h
@@ -13,6 +13,7 @@
 #include "nsIObserverService.h"
 #include "nsIObserver.h"
 #include "nsString.h"
+#include "nsWidgetsCID.h"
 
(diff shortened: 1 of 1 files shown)


rainemak:
Can this be done other way round so that embedlite Clipboard aligns against upstream so that we do not need patching?

llewelld:
That would make more sense. I'll take a look at doing this in the morning.

llewelld:
Thanks for the suggestion @rainemak, this is definitely a cleaner approach.

okodron (approved):
Looks better now.

rainemak (approved):
Looks indeed better.

PR #129, opened 2021-12-17
Ensure zooming is always allowed if desktop mode is active. Fixes JB#56767 OMP#JOLLA-609

llewelld:
If desktop mode is active then we should allow page zoomingi by the user (pinch-to-zoom), even if the metadata for the page has "user-scalable=no" set.


llewelld:
This is a much smaller change than the messy diff would imply.

okodron (approved):

rainemak (approved):
Didn't test on device but change looks good. Hence, approving.

rainemak:
Noticed minor typo in the commit message: "zoomingi" => "zooming"

llewelld:
Thanks; typo fixed.

paveltkv (approved):

PR #130, opened 2021-12-17
Pass double taps to embedhelper.js. JB#56765 OMP#JOLLA-607

llewelld:
Ensure that double taps are passed on to the embedhelper.js, so they can be passed on to any javascript handlers.

Set the decision about whether to zoom on double tap follow the decision made by nsLayoutUtils::AllowZoomingForDocument().

### DIFF
--- a/embedding/embedlite/embedding.js
+++ b/embedding/embedlite/embedding.js
@@ -111,7 +111,6 @@ pref("gfx.xrender.enabled", false);
 // AZPC overrides, see EmbedLiteViewChild.cpp
 pref("embedlite.azpc.handle.viewport", true);
 pref("embedlite.azpc.handle.singletap", false);
-pref("embedlite.azpc.handle.doubletap", true);
 pref("embedlite.azpc.handle.longtap", false);
 pref("embedlite.azpc.handle.scroll", true);
 pref("embedlite.azpc.json.viewport", true);
--- a/embedding/embedlite/embedding.js
+++ b/embedding/embedlite/embedding.js
@@ -107,15 +107,19 @@ pref("media.prefer-gstreamer", true);
 pref("media.gstreamer.enable-blacklist", false);
 // Disable X backend on GTK
 pref("gfx.xrender.enabled", false);
+
+// AZPC overrides, see EmbedLiteViewChild.cpp
 pref("embedlite.azpc.handle.viewport", true);
-pref("embedlite.azpc.handle.singletap", true);
-pref("embedlite.azpc.handle.longtap", true);
+pref("embedlite.azpc.handle.singletap", false);
+pref("embedlite.azpc.handle.doubletap", true);
+pref("embedlite.azpc.handle.longtap", false);
 pref("embedlite.azpc.handle.scroll", true);
-pref("embedlite.azpc.json.viewport", false);
-pref("embedlite.azpc.json.singletap", false);
-pref("embedlite.azpc.json.doubletap", false);
-pref("embedlite.azpc.json.longtap", false);
+pref("embedlite.azpc.json.viewport", true);
+pref("embedlite.azpc.json.singletap", true);
+pref("embedlite.azpc.json.doubletap", true);
+pref("embedlite.azpc.json.longtap", true);
 pref("embedlite.azpc.json.scroll", false);
+
 // Make gecko compositor use GL context/surface provided by the application.
 pref("embedlite.compositor.external_gl_context", false);
 // Request the application to create GLContext for the compositor as
--- a/embedding/embedlite/embedshared/EmbedLiteViewChild.cpp
+++ b/embedding/embedlite/embedshared/EmbedLiteViewChild.cpp
@@ -70,7 +70,6 @@ static struct {
     bool viewport;
     bool scroll;
     bool singleTap;
-    bool doubleTap;
     bool longTap;
 } sHandleDefaultAZPC;
 static struct {
@@ -88,7 +87,6 @@ static void ReadAZPCPrefs()
   // Init default azpc notifications behavior
   Preferences::AddBoolVarCache(&sHandleDefaultAZPC.viewport, "embedlite.azpc.handle.viewport", true);
   Preferences::AddBoolVarCache(&sHandleDefaultAZPC.singleTap, "embedlite.azpc.handle.singletap", false);
-  Preferences::AddBoolVarCache(&sHandleDefaultAZPC.doubleTap, "embedlite.azpc.handle.doubletap", true);
   Preferences::AddBoolVarCache(&sHandleDefaultAZPC.longTap, "embedlite.azpc.handle.longtap", false);
   Preferences::AddBoolVarCache(&sHandleDefaultAZPC.scroll, "embedlite.azpc.handle.scroll", true);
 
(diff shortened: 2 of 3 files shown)


okodron (approved):

okodron (approved):

mkenttala (approved):

embedding/embedlite/embedshared/EmbedLiteViewChild.cpp
```diff
@@ -979,13 +979,15 @@ mozilla::ipc::IPCResult EmbedLiteViewChild::RecvHandleDoubleTap(const LayoutDevi
...
-      document->GetDocumentElement(), &presShellId, &viewId)) {
-    ZoomToRect(presShellId, viewId, zoomToRect);
+  if (nsLayoutUtils::AllowZoomingForDocument(document)) {
+    CSSRect zoomToRect = CalculateRectToZoomTo(document, cssPoint);
+    uint32_t presShellId;
+    ViewID viewId;
+    if (APZCCallbackHelper::GetOrCreateScrollIdentifiers(
+        document->GetDocumentElement(), &presShellId, &viewId)) {
+      LOGW("TAP: ZoomeToRect");
+      ZoomToRect(presShellId, viewId, zoomToRect);
```
rainemak, line 989: Out of curiosity, that's the reason why GetApzcTreeManager()->ZoomToRect cannot cope with this? So, going down here AsyncPanZoomController::ZoomToRect

llewelld:
Removed an extraneous debug print that wasn't supposed to be there.

embedding/embedlite/embedshared/EmbedLiteViewChild.cpp
llewelld: I'm probably misreading the code, and/or misunderstanding your comment, but doesn't this code end up doing exactly that? So, the inverted call stack here goes:
1. EmbedLiteViewChild::ZoomToRect()
2. PEmbedLiteVeiwChild::SendZoomToRect()
3. EmbedLiteViewParent::RecvZoomToRect()
4. ApzcTreeManager::ZoomToRect()
5. AsyncPanZoomController::ZoomToRect()

rainemak: Correct

rainemak: I just pondering why apzc is not triggering the zoomtherect

embedding/embedlite/embedshared/EmbedLiteViewChild.cpp
```diff
@@ -979,13 +979,14 @@ mozilla::ipc::IPCResult EmbedLiteViewChild::RecvHandleDoubleTap(const LayoutDevi
...
+    CSSRect zoomToRect = CalculateRectToZoomTo(document, cssPoint);
+    uint32_t presShellId;
+    ViewID viewId;
+    if (APZCCallbackHelper::GetOrCreateScrollIdentifiers(
+        document->GetDocumentElement(), &presShellId, &viewId)) {
+      ZoomToRect(presShellId, viewId, zoomToRect);
+    }
   }
 
   if (sPostAZPCAsJson.doubleTap) {
```
rainemak, line 992: You should enable this to the embedding.js in order this pipeline to work. Currently, it is not enabled.

When enabled Gesture:DoubleTap starts interfering with doubletaptozoom. For instance visible in the testinputfocus.html (zooming allowed) when do following:

1) Open test case
2) Pinch in so that last three paragraphs do fit to the screen
3) Double tap the last paragraph

Expected outcome:
Page is fully zoomed out

Actual outcome:
Page is fully zoomed out and text gets selected

Additional info:
Easier to reproduce with a page that does not have contentEditable set.

rainemak: Would be good to move defaults that we have in webenginesettings to the embedding.js

rainemak (changes requested):
Currently this doesn't yet work properly even if Gesture:DoubleTap is enabled. Out of my head one approach could be to trigger before ```nsLayoutUtils::AllowZoomingForDocument(document)``` a hittest event and based on result either trigger zoom to rect or Gesture:DoubleTap.

embedding/embedlite/embedshared/EmbedLiteViewChild.cpp
```diff
@@ -979,16 +1005,54 @@ mozilla::ipc::IPCResult EmbedLiteViewChild::RecvHandleDoubleTap(const LayoutDevi
...
+
+      nsIScrollableFrame* rootScrollFrame =
+          presShell->GetRootScrollFrameAsScrollable();
+      if (rootScrollFrame) {
+        FrameMetrics metrics =
+            nsLayoutUtils::CalculateBasicFrameMetrics(rootScrollFrame);
+        CSSRect compositedArea(
+            CSSPoint::FromAppUnits(presShell->GetVisualViewportOffset()),
+            metrics.CalculateCompositedSizeInCssPixels());
+      }
```
rainemak, line 1051: Is this if-block a leftover?

embedding/embedlite/embedding.js
```diff
@@ -107,15 +107,19 @@ pref("media.prefer-gstreamer", true);
 pref("media.gstreamer.enable-blacklist", false);
 // Disable X backend on GTK
 pref("gfx.xrender.enabled", false);
+
+// AZPC overrides, see EmbedLiteViewChild.cpp
 pref("embedlite.azpc.handle.viewport", true);
-pref("embedlite.azpc.handle.singletap", true);
-pref("embedlite.azpc.handle.longtap", true);
+pref("embedlite.azpc.handle.singletap", false);
+pref("embedlite.azpc.handle.doubletap", true);
```
rainemak, line 114: We could consider dropping this one altogether as ```sHandleDefaultAZPC.doubleTap``` is not referenced (i.e. embedlite.azpc.handle.doubletap doesn't do anything.

As a counterpart do we need that ```embedlite.azpc.json.doubletap``` in reality either. We could have hittesting done always.

llewelld:
This relies on:
https://github.com/sailfishos/embedlite-components/pull/90
https://github.com/sailfishos/sailfish-components-webview/pull/146

Test case is here:
https://github.com/sailfishos/sailfish-browser/pull/967

Temporary page for testing:
https://www.flypig.co.uk/tests/dblclick.html

embedding/embedlite/embedshared/EmbedLiteViewChild.cpp
llewelld: Yes, old debugging code, now removed.

embedding/embedlite/embedding.js
llewelld: Right, the former is a dead setting. I can remove it.

For `embedlite.azpc.json.doubletap`: setting this to false will bring the browser behaviour more in line with Android FF (which, surprisingly to me, always seems to override `dblclick` handlers with zoom), so maybe it's worth keeping for that reason.

But I'm not really certain what the broader semantics of these settings are intended for. Perhaps this is the wrong setting for that?

rainemak: Let's drop the dead one.

embedding/embedlite/embedshared/EmbedLiteViewChild.cpp
```diff
@@ -69,7 +70,7 @@ static struct {
     bool viewport;
     bool scroll;
     bool singleTap;
-    bool doubleTap;
+    // bool doubleTap; // Unused
```
okodron, line 73: Can we drop this line?

llewelld: Sure; removed.

rainemak (approved):
Works nicely now! Much better that the previous version. Thanks @llewelld

rainemak:
Rebase before merging.

llewelld:
Rebased.

PR #131, opened 2021-12-20
[sailfishos][embedlite] Load external uris directory via nsIExternalProtocolService. Fixes JB#56648 OMP#JOLLA-577

rainemak:
The nsDocShell can trigger directly external protocol service regardless
of the anchor target without creating new tab. Exposed protocols are only
checked by the nsDocShell. Hence, this ought to be safe.

### DIFF
--- a/embedding/embedlite/embedding.js
+++ b/embedding/embedlite/embedding.js
@@ -169,8 +169,14 @@ pref("network.protocol-handler.external-default", true);      // OK to load
 pref("network.protocol-handler.warn-external-default", false); // Do not warn
 
 // This pref controls the default settings.  Per protocol settings can be used
-// to override this value.
+// to override this value. See nsDocShell::OnLinkClickSync and
+// nsExternalHelperAppService::IsExposedProtocol. Protocol preference
+// takes precedence.
 pref("network.protocol-handler.expose-all", true);
+pref("network.protocol-handler.expose.sms", false);
+pref("network.protocol-handler.expose.mailto", false);
+pref("network.protocol-handler.expose.tel", false);
+pref("network.protocol-handler.expose.geo", false);
 
 /* disable some protocol warnings */
 pref("network.protocol-handler.warn-external.tel", false);


attah:
Do you mean _ought_ to be safe? 🙃

rainemak:
> Do you mean _ought_ to be safe? upside_down_face

Tampering anything that touches nsDocShell is potentially error prone or changes behaviour of it as well. Each of this kind errors take long to figure out (even for me). Here are my findings:

I checked if there would be easy way to get onlclicked registered in WebBrowserChrome. That didn't work out. As the scheme is not http we're not getting this as a nsIHttpChannel either. Went through backtrace frames and checking frames from where external urls are getting triggered. 

Finally realized that external urls can be marked as not exposed to web content and to be handled directly by the external app. Assumption of the PR is that mailto, geo, sms, tel are each handled by an external application. See nsDocShell::OnLinkClickSync line 11880 and nsExternalHelperAppService::IsExposedProtocol line 915.

attah:
It was not a technical comment, just a nitpick on the wording. Sorry for the noise.

embedding/embedlite/embedding.js
```diff
@@ -169,8 +169,14 @@ pref("network.protocol-handler.external-default", true);      // OK to load
 pref("network.protocol-handler.warn-external-default", false); // Do not warn
 
 // This pref controls the default settings.  Per protocol settings can be used
-// to override this value.
+// to override this value. See nsDocShell::OnLinkClickSync and
+// nsExternalHelperAppService::IsExposedProtocol. Protocol preference
+// takes precedence.
 pref("network.protocol-handler.expose-all", true);
+pref("network.protocol-handler.expose.sms", false);
+pref("network.protocol-handler.expose-mailto", false);
```
attah, line 177: Should it be .mailto etc?

rainemak:
> It was not a technical comment, just a nitpick on the wording. Sorry for the noise.

Fixed that nitpick already.

attah (commented):
The way i read http://kb.mozillazine.org/Network.protocol-handler.expose-all and http://kb.mozillazine.org/Network.protocol-handler.expose.(protocol), the -all version overrides the others (regardless of order). But i know nothing, and info may be old.

rainemak:
I hope that my explanation was good from tech point of view. I missed that from original comment.

embedding/embedlite/embedding.js
rainemak: Good catch! Copy paste error.

mkenttala (approved):

okodron (approved):

PR #135, opened 2022-01-05
[sailfishos][gecko] Fix memory reporting of wasm memory. Fixes JB#56908

llewelld:
Backported from upstream gecko 94, bug 1615988.

This commit makes several tweaks to memory reporting for wasm memory.
  * Add a field for shared wasm memory and track it in SAB.
    - Before this commit, shared wasm memory wouldn't report its guard
      pages but should.
  * Track wasm guard pages in RuntimeSizes, not ClassInfo
    - We want wasm guard pages to show as a top-level memory report
      item, similar to vmem, and not under the owning array object.
      Displaying under the owning object bloats the tree of memory usage
      with large amounts of memory that's only reserved and not
      committed, which is confusing. Before this commit the class info
      reporter would try to make this happen, but this approach was
      broken and the simplest fix is to
      track this on RuntimeSizes and report the value from their.
  * Only add wasm memory if the buffer is attached. Detached buffers may
    still be live but no longer own the wasm heap and shouldn't report
    their old size.

Differential Revision: https://phabricator.services.mozilla.com/D124390

### DIFF
--- a/rpm/0087-sailfishos-gecko-Fix-memory-reporting-of-wasm-memory.patch
+++ b/rpm/0087-sailfishos-gecko-Fix-memory-reporting-of-wasm-memory.patch
@@ -1,0 +1,279 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Ryan Hunt <rhunt@eqrion.net>
+Date: Tue, 7 Sep 2021 21:03:48 +0000
+Subject: [PATCH] [sailfishos][gecko] Fix memory reporting of wasm memory.
+ Fixes JB#56908
+
+Backported from upstream gecko 94, bug 1615988.
+
+This commit makes several tweaks to memory reporting for wasm memory.
+  * Add a field for shared wasm memory and track it in SAB.
+    - Before this commit, shared wasm memory wouldn't report its guard pages but should.
+  * Track wasm guard pages in RuntimeSizes, not ClassInfo
+    - We want wasm guard pages to show as a top-level memory report item, similar to vmem,
+      and not under the owning array object. Displaying under the owning object bloats the
+      tree of memory usage with large amounts of memory that's only reserved and not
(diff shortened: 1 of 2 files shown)


llewelld:
The original patch for reference: https://hg.mozilla.org/mozilla-central/rev/1862f870e642

rpm/0087-sailfishos-gecko-Fix-memory-reporting-of-wasm-memory.patch
```diff
@@ -0,0 +1,276 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Ryan Hunt <rhunt@eqrion.net>
+Date: Tue, 7 Sep 2021 21:03:48 +0000
+Subject: [PATCH] [sailfishos][gecko] Fix memory reporting of wasm memory.
+ Fixes JB#56908
+
+Backported from upstream gecko 94, bug 1615988.
```
rainemak, line 7: Please add link to the https://github.com/sailfishos-mirror/gecko-dev/commit/c871f743159fbbf2e1b22d877ace761360d72c03 as well.

rainemak (approved):
Adding a small note regarding git mirror linking.

rpm/0087-sailfishos-gecko-Fix-memory-reporting-of-wasm-memory.patch
rainemak: Please add this git mirror link to the git commit message as well (https://github.com/sailfishos/gecko-dev/pull/135/commits/971ade0fc05e78d16740565a9e91c3daa4ff6baf)

llewelld: Thanks, I've added it to both the commit summary and the patch description.

PR #136, opened 2022-01-26
[sailfishos][gecko] Do not remove debugging symbols

d-grigorev:
This patch disables the generation of debug information for rust and
returns the contents of the debuginfo package.

Add 'a' character to the milestone in rpm/xulrunner-qt5.spec to build
an alpha version of the engine.

### DIFF
--- a/rpm/xulrunner-qt5.spec
+++ b/rpm/xulrunner-qt5.spec
@@ -20,8 +20,8 @@
 %define system_libwebp      1
 
 
-%global mozappdir     %{_libdir}/%{name}-%{milestone}
-%global mozappdirdev  %{_libdir}/%{name}-devel-%{milestone}
+%global mozappdir     %{_libdir}/%{name}-%{greversion}
+%global mozappdirdev  %{_libdir}/%{name}-devel-%{greversion}
 
 # Private/bundled libs the final package should not provide or depend on.
 %global privlibs             libfreebl3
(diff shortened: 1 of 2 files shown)


rpm/xulrunner-qt5.spec
```diff
@@ -20,8 +20,8 @@
 %define system_libwebp      1
 
 
-%global mozappdir     %{_libdir}/%{name}-%{milestone}
-%global mozappdirdev  %{_libdir}/%{name}-devel-%{milestone}
+%global mozappdir     %{_libdir}/%{name}-%{greversion}
+%global mozappdirdev  %{_libdir}/%{name}-devel-%{greversion}
```
llewelld, line 24: Is the idea here to ensure the app directories don't change when running an alpha version?

llewelld (commented):
This built okay for me (i486 and aarch64 targets on OBS), and would make a really nice changes, so I'm in favour of merging it at this point, after a rebase.

rpm/xulrunner-qt5.spec
d-grigorev: I tried to append 'a' to the milestone and got packaging issue, so `%{_libdir}/%{name}-devel-%{greversion}` was correct for me. I can't claim that the build system really creates an alpha build. Will test it and come back with feedback.

llewelld: Thanks for explaining. In normal use it looks like this change has no effect, so if that works for you, then I'm happy with it.

d-grigorev: I confirm that the clean build with extra 'a' character in `%{milestone}` succeeds.

d-grigorev:
Rebased.

PR #139, opened 2022-02-18
[sailfishos][embedlite] Prefer profile directory for search engines. Fixes JB#57275

rainemak:
Only if profile directory is not available, use hardcoded path.
Currently all browser downloaded search engines are leaking to
all other apps using engine.

This is last change to get tst_searchengine of QtMozEmbed properly with local runs. When running locally test case saw also browser installed search engines and only way to avoid that was to remove browser installed search engines.


mkenttala (approved):

llewelld (approved):
LGTM.

PR #143, opened 2022-03-10
[sailfishos][gecko] Initialise SVGGeometryProperty::ResolveAll parameters. JB#56892

llewelld:
During SVG rendering HasValidDimensions() was sporadically returning false, causing some SVG elements not to render correctly.

The problem appears to be an optimisation glitch. This change initialises the parameters passed by reference to SVGGeometryProperty::ResolveAll, so that the compiler doesn't determine that they may be undefined.

### DIFF
--- a/rpm/0091-sailfishos-gecko-Initialise-SVGGeometryProperty-Reso.patch
+++ b/rpm/0091-sailfishos-gecko-Initialise-SVGGeometryProperty-Reso.patch
@@ -1,0 +1,246 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: David Llewellyn-Jones <david.llewellyn-jones@jolla.com>
+Date: Wed, 9 Mar 2022 22:16:56 +0000
+Subject: [PATCH] [sailfishos][gecko] Initialise
+ SVGGeometryProperty::ResolveAll parameters. JB#56892
+
+During SVG rendering HasValidDimensions() was sporadically returning
+false, causing some SVG elements not to render correctly.
+
+The problem appears to be an optimisation glitch. This change
+initialises the parameters passed by reference to
+SVGGeometryProperty::ResolveAll, so that the compiler doesn't determine
+that they may be undefined.
+---
+ dom/svg/SVGCircleElement.cpp           | 4 +++-
+ dom/svg/SVGEllipseElement.cpp          | 3 +++
+ dom/svg/SVGForeignObjectElement.cpp    | 2 ++
+ dom/svg/SVGImageElement.cpp            | 2 ++
+ dom/svg/SVGRectElement.cpp             | 5 +++++
+ layout/svg/nsSVGForeignObjectFrame.cpp | 4 ++++
+ layout/svg/nsSVGImageFrame.cpp         | 6 ++++++
+ 7 files changed, 25 insertions(+), 1 deletion(-)
+
+diff --git a/dom/svg/SVGCircleElement.cpp b/dom/svg/SVGCircleElement.cpp
+index 6fc8d29c2e7c..bba2a93b9a65 100644
+--- a/dom/svg/SVGCircleElement.cpp
++++ b/dom/svg/SVGCircleElement.cpp
+@@ -70,7 +70,7 @@ already_AddRefed<DOMSVGAnimatedLength> SVGCircleElement::R() {
+ 
+ /* virtual */
+ bool SVGCircleElement::HasValidDimensions() const {
+-  float r;
++  float r = 1.0f;
+ 
+   MOZ_ASSERT(GetPrimaryFrame());
+   SVGGeometryProperty::ResolveAll<SVGT::R>(this, &r);
+@@ -89,6 +89,7 @@ bool SVGCircleElement::GetGeometryBounds(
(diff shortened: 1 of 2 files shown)


rpm/0091-sailfishos-gecko-Initialise-SVGGeometryProperty-Reso.patch
```diff
@@ -0,0 +1,246 @@
...
+diff --git a/dom/svg/SVGCircleElement.cpp b/dom/svg/SVGCircleElement.cpp
+index 6fc8d29c2e7c..bba2a93b9a65 100644
+--- a/dom/svg/SVGCircleElement.cpp
++++ b/dom/svg/SVGCircleElement.cpp
+@@ -70,7 +70,7 @@ already_AddRefed<DOMSVGAnimatedLength> SVGCircleElement::R() {
+ 
+ /* virtual */
+ bool SVGCircleElement::HasValidDimensions() const {
+-  float r;
++  float r = 1;
```
rainemak, line 33: Why r = 1; instead of r = 0.0f;

rainemak, line 33: Why r = 1; instead of r = 0.0f;

rpm/0091-sailfishos-gecko-Initialise-SVGGeometryProperty-Reso.patch
```diff
@@ -0,0 +1,246 @@
...
+   float x, y;
++  x = y = 0.0f;
+ 
+   if (GetPrimaryFrame()) {
+     SVGGeometryProperty::ResolveAll<SVGT::X, SVGT::Y>(this, &x, &y);
+@@ -101,6 +102,7 @@ gfxMatrix SVGForeignObjectElement::PrependLocalTransformsTo(
+ /* virtual */
+ bool SVGForeignObjectElement::HasValidDimensions() const {
+   float width, height;
++  width = height = 1.0f;
```
rainemak, line 97: similar to line https://github.com/sailfishos/gecko-dev/pull/143/files#r823957754

rainemak, line 97: similar to line https://github.com/sailfishos/gecko-dev/pull/143/files#r823957754

rpm/0091-sailfishos-gecko-Initialise-SVGGeometryProperty-Reso.patch
```diff
@@ -0,0 +1,246 @@
...
+   Rect rect;
++  rect.x = rect.y = rect.width = rect.height = 1.0f;
+ 
+   MOZ_ASSERT(GetPrimaryFrame());
+   SVGGeometryProperty::ResolveAll<SVGT::X, SVGT::Y, SVGT::Width, SVGT::Height>(
+@@ -278,6 +279,7 @@ already_AddRefed<Path> SVGImageElement::BuildPath(PathBuilder* aBuilder) {
+ /* virtual */
+ bool SVGImageElement::HasValidDimensions() const {
+   float width, height;
++  width = height = 1.0f;
```
rainemak, line 117: Cannot quite get why some floats that are passed to ResolveAll are initialized to 0.0f and some 1.0f. Could @llewelld please elaborate?

rainemak:
Added comments / questions. The SVG rendering issue on 64bit platform seems to be gone which is nice! Clearly points towards random memory like you explained.

rainemak:
Added comments / questions. The SVG rendering issue on 64bit platform seems to be gone which is nice! Clearly points towards random memory like you explained.

rpm/0091-sailfishos-gecko-Initialise-SVGGeometryProperty-Reso.patch
llewelld: It's so that, in the event that `r` isn't touched by the call to `ResolveAll`, the condition that follows defaults to `true` and the object is rendered by default:
```
  return r > 0;
```
Since I didn't dig deep enough to know whether the problem is due to the value being left uninitialised, or an incorrect optimisation due to the compiler only *thinking* that it's uninitialised, I thought it better to err on the side of caution and cover both possibilities.

You're right it should at least be `r = 1.0f;` though; I'll fix this.

rpm/0091-sailfishos-gecko-Initialise-SVGGeometryProperty-Reso.patch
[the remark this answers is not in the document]
llewelld: To add to my other comment, I set them to 1 only where I saw a clear possibility that the condition should return true if the value were left uninitialised. In most cases, were I didn't seen any obvious path to a condition, I set them to 0.

rpm/0091-sailfishos-gecko-Initialise-SVGGeometryProperty-Reso.patch
rainemak:
```suggestion
+  float r = 1.0f;
```
So, 1.0f as well.

rainemak (approved):
Very good, thank you @llewelld

PR #145, opened 2022-03-22
[sailfishos][embedlite] Add a content controller stack. Fixes JB#57690

llewelld:
When multiple WebView componennts are used in an app, since there is
only one compositor, only one can be in direct use at a time. This will
be the last that was set as active.

When the currently activated component is deactivated, the WebView that
was active prior to that should be reinstanted. In particular, the
view's GeckoContentController needs to be reinstated.

This change adds a stack for managing the content controllers, so that
previously active controllers can be reactivated.

### DIFF
--- a/embedding/embedlite/embedshared/EmbedLiteViewParent.cpp
+++ b/embedding/embedlite/embedshared/EmbedLiteViewParent.cpp
@@ -67,6 +67,10 @@ EmbedLiteViewParent::~EmbedLiteViewParent()
 {
   MOZ_COUNT_DTOR(EmbedLiteViewParent);
   LOGT("mView:%p, mCompositor:%p", mView, mCompositor.get());
+  nsWindow *window = GetWindowWidget();
+  if (window) {
+    window->Deactivate(mContentController);
+  }
   mContentController = nullptr;
   mWindow.RemoveObserver(this);
 }
@@ -75,6 +79,10 @@ void
 EmbedLiteViewParent::ActorDestroy(ActorDestroyReason aWhy)
 {
   LOGT("reason: %i", aWhy);
+  nsWindow *window = GetWindowWidget();
+  if (window) {
+    window->Deactivate(mContentController);
+  }
   mContentController = nullptr;
 }
 
@@ -101,7 +109,7 @@ EmbedLiteViewParent::UpdateScrollController()
 
     nsWindow *window = GetWindowWidget();
     if (window) {
-      window->SetContentController(mContentController);
+      window->Activate(mContentController);
     }
   }
 }
@@ -639,6 +647,24 @@ EmbedLiteViewParent::SetEmbedAPIView(EmbedLiteView *aView)
 {
   LOGT();
   mView = aView;
+  return NS_OK;
+}
+
+NS_IMETHODIMP
+EmbedLiteViewParent::SetIsActive(bool aIsActive)
+{
+  nsWindow *window = GetWindowWidget();
+
+  if (window) {
+    if (aIsActive) {
+      if (mCompositor) {
+        window->Activate(mContentController);
+      }
+    } else {
+      window->Deactivate(mContentController);
+    }
+  }
+
   return NS_OK;
 }
 
--- a/embedding/embedlite/embedshared/nsWindow.cpp
+++ b/embedding/embedlite/embedshared/nsWindow.cpp
@@ -34,6 +34,15 @@ using namespace mozilla::layers;
 using namespace mozilla::widget;
 //using namespace mozilla::ipc;
 
+#ifdef MOZ_LOGGING
+#define LOG_CONTROLLER_STACK  \
+  if (MOZ_UNLIKELY(detail::log_test(GetEmbedCommonLog("EmbedLiteTrace"), LogLevel::Debug))) { \
+    LogControllerStack(mControllers); \
+  }
+#else // MOZ_LOGGING
+#define LOG_CONTROLLER_STACK  do {} while (0)
+#endif // MOZ_LOGGING
+
 namespace mozilla {
 namespace embedlite {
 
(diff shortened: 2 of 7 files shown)


embedding/embedlite/embedshared/nsWindow.cpp
```diff
@@ -276,11 +285,57 @@ layers::LayersId nsWindow::GetRootLayerId() const
...
+  }
+}
+
+void nsWindow::DropContentController(EmbedContentController* aController)
+{
+  if (!aController || mControllers.empty()) {
+    return;
+  }
+
+  if (aController == mControllers.back()) {
```
rainemak, line 322: Better compare against controller unique id and this way safe guarding against plausible invalid pointer comparison.

embedding/embedlite/embedshared/nsWindow.cpp
```diff
@@ -276,11 +285,57 @@ layers::LayersId nsWindow::GetRootLayerId() const
...
+    if (mCompositorSession && !mControllers.empty()) {
+      mCompositorSession->SetContentController(mControllers.back());
+    }
+  }
+  else {
+    // An inactive view deactivated, so we erase all instances of its
+    // content controllers from the stack
+    LOGT("Erasing content controller %u", aController->GetUniqueID());
+    mControllers.erase(std::remove(mControllers.begin(), mControllers.end(),
+                                   aController), mControllers.end());
```
rainemak, line 336: Over here as well predicate should be comparing unique ids.

embedding/embedlite/embedshared/EmbedLiteViewParent.cpp
```diff
@@ -642,5 +650,25 @@ EmbedLiteViewParent::SetEmbedAPIView(EmbedLiteView *aView)
...
+{
+  nsWindow *window = GetWindowWidget();
+
+  if (window) {
+    if (aIsActive) {
+      if (mCompositor) {
+        window->SetContentController(mContentController);
+      }
+    }
+    else {
```
rainemak, line 664:
```suggestion
    } else {
```

embedding/embedlite/embedshared/EmbedLiteViewParent.cpp
```diff
@@ -642,5 +650,25 @@ EmbedLiteViewParent::SetEmbedAPIView(EmbedLiteView *aView)
...
+      window->DropContentController(mContentController);
+    }
+  }
+
+  return NS_OK;
+}
+
```
rainemak, line 671: Extra empty line

embedding/embedlite/embedshared/nsWindow.cpp
```diff
@@ -276,11 +285,57 @@ layers::LayersId nsWindow::GetRootLayerId() const
...
 {
   if (mCompositorSession) {
     mCompositorSession->SetContentController(aController);
   }
+
+  MOZ_ASSERT(aController);
+
+  // If the view is deactivated we need to reactivate the previous view's
+  // content controller, so we store the controllwers on a stack
+  if (mControllers.empty() || aController != mControllers.back()) {
```
rainemak, line 309:
```suggestion
  if (mControllers.empty() || aController->GetUniqueID() != mControllers.back()->GetUniqueID()) {
```

Could it happen that we have multiple instances of the same content controller in the stack? Wondering if we should always move the existing instance to the top. Surely, that is not needed if there cannot be multiple instances.

embedding/embedlite/embedshared/nsWindow.cpp
```diff
@@ -276,11 +285,57 @@ layers::LayersId nsWindow::GetRootLayerId() const
...
+    LOGT("Pushing content controller %u", aController->GetUniqueID());
+    mControllers.push_back(aController);
+    LOG_CONTROLLER_STACK;
+  }
+}
+
+void nsWindow::DropContentController(EmbedContentController* aController)
```
rainemak, line 316: Wouldn't it be sufficient to have this called like ```nsWindow::Release(EmbedContentController *aController)```, would it? Maybe this could even be Release & Reserve pair. Somehow I find this repetition of ContentController unnecessary.

And yes like ```(EmbedContentController *aController)``` => there's a mixture of different styles in the embedlite we should be heading towards this style.

embedding/embedlite/embedshared/nsWindow.cpp
```diff
@@ -276,11 +285,57 @@ layers::LayersId nsWindow::GetRootLayerId() const
...
+{
+  int pos = 0;
+  for (std::list<EmbedContentController *>::const_iterator it = controllers.begin(); it != controllers.end(); ++it) {
+    LOGT("Stack controller %d : %u", pos, (*it)->GetUniqueID());
+    ++pos;
+  }
+}
+#endif // MOZ_LOGGING
+
+void nsWindow::SetContentController(EmbedContentController *aController)
```
rainemak, line 299: How does nsWindow::Reserve(EmbedContentController *aController) or nsWindow::Activate(EmbedContentController *aController) sound like?

Maybe this could Reserve (SetContentController) & Release (DropContentController) or Activate (SetContentController) & Deactivate (DropContentController) pair. Now that I said Activate & Deactivate that pair sounds more natural to me. What do you think?

embedding/embedlite/embedshared/nsWindow.h
```diff
@@ -88,7 +90,8 @@ class nsWindow : public PuppetWidgetBase
   uint32_t GetUniqueID() const;
   layers::LayersId GetRootLayerId() const;
 
-  void SetContentController(mozilla::layers::GeckoContentController* aController);
+  void SetContentController(EmbedContentController* aController);
+  void DropContentController(EmbedContentController* aController);
```
rainemak, line 94: And this `Deactivate`

embedding/embedlite/embedshared/nsWindow.cpp
```diff
@@ -276,11 +285,65 @@ layers::LayersId nsWindow::GetRootLayerId() const
...
+  // If the view is deactivated we need to reactivate the previous view's
+  // content controller, so we store the controllwers on a stack
+  if (mControllers.empty() || !ControllersEqual(aController, mControllers.back())) {
+    LOGT("Pushing content controller %u", aController->GetUniqueID());
+    mControllers.push_back(aController);
+    LOG_CONTROLLER_STACK;
+  }
+}
+
+void nsWindow::Deactivate(EmbedContentController* aController)
```
rainemak, line 323:
```suggestion
void nsWindow::Deactivate(EmbedContentController *aController)
```

embedding/embedlite/embedshared/nsWindow.cpp
```diff
@@ -276,11 +285,65 @@ layers::LayersId nsWindow::GetRootLayerId() const
...
+    mControllers.pop_back();
+    if (mCompositorSession && !mControllers.empty()) {
+      mCompositorSession->SetContentController(mControllers.back());
+    }
+  } else {
+    // An inactive view deactivated, so we erase all instances of its
+    // content controllers from the stack
+    LOGT("Erasing content controller %u", aController->GetUniqueID());
+    mControllers.erase(std::remove_if(mControllers.begin(), mControllers.end(),
+                                      [aController](const EmbedContentController* aOther) {
```
rainemak, line 342:
```suggestion
                                      [aController](const EmbedContentController *aOther) {
```

embedding/embedlite/embedshared/nsWindow.cpp
```diff
@@ -276,11 +285,65 @@ layers::LayersId nsWindow::GetRootLayerId() const
...
+    ++pos;
+  }
+}
+#endif // MOZ_LOGGING
+
+inline static bool ControllersEqual(const EmbedContentController *aController1, const EmbedContentController *aController2)
+{
+  // Compare unique ids, but fallback to pointer comparison if id is zero
+  return (aController1->GetUniqueID() != 0 && (aController1->GetUniqueID() == aController2->GetUniqueID()))
+      || (aController1 == aController2);
```
rainemak, line 303: What's reason for the fallback? If unique ids are not matching then this should return false, right?

llewelld:
Squashed and rebased.

embedding/embedlite/embedshared/nsWindow.cpp
rainemak: Let's leave it like it is.

rainemak (approved):

PR #147, opened 2022-04-04
[sailfishos][embedlite] Drop radio and checkbox max size restriction. Fixes JB#57693

llewelld:
Remove the restriction on the maximum size of radio and checkbox input elements, so that they can be made larger than the default.

This is to match the behaviour of desktop and Android Firefox.

In particular, this fixes a but that prevented [Material toggle button components](https://material-components.github.io/material-web/demos/switch) from working correctly.

### DIFF
--- a/embedding/embedlite/tests/content/embedScrollStyles.css
+++ b/embedding/embedlite/tests/content/embedScrollStyles.css
@@ -171,8 +171,6 @@ button {
 
 input[type="radio"],
 input[type="checkbox"] {
-  max-width: 14px;
-  max-height: 14px;
   border: 1px solid #a7a7a7 !important;
   padding: 2px 1px 2px 1px;
 }


llewelld:
This file ends up as ./chrome/toolkit/content/global/embedScrollStyles.css inside omni.ja and can be edited directly using the instructions in the [browser documentation](https://docs.sailfishos.org/Reference/Core_Areas_and_APIs/Browser/Working_with_Browser/#debugging-omnija-of-gecko).

I was a bit surprised that a CSS file inside a `test` directory gets used for the production styling. I didn't look into the history or reason, but I wonder if that's something we should consider restructuring at some point?

The screenshot below shows the effect of applying this change when using the Material toggle button.

checkbox

llewelld:
@piggz: I can't add you directly as a reviewer, but please feel free to comment on this change.

mkenttala (approved):

rainemak (approved):
In any case this could go. Don't worry about the history of the embedScrollStyle.css location.

piggz:
all i can add is that this fixes home-assistant web ui for me, which uses the material components, good find

llewelld:
Thanks @piggz for the feedback, and also for raising this in the first place.

PR #148, opened 2022-08-18
[sailfishos][embedlite] Clean background gradients from input elements. Fixes JB#58457

rainemak:
This covers textarea, input, select, and button.

Instructions regarding [omni debugging](https://docs.sailfishos.org/Reference/Core_Areas_and_APIs/Browser/Working_with_Browser/#debugging-omnija-of-gecko) helps when testing this.

### DIFF
--- a/embedding/embedlite/tests/content/embedScrollStyles.css
+++ b/embedding/embedlite/tests/content/embedScrollStyles.css
@@ -110,7 +110,7 @@ textarea,
   border-style: solid;
   border-color: #7d7d7d;
   color: #414141;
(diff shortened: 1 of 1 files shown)


llewelld (approved):
Works nicely.

For me, the default text and border of the buttons etc. is rather thin, making them harder to notice now they're no longer embossed (previously they stood out due to the backgrounds). So maybe there's an argument for using stronger lines or font.

But that's a separate discussion as this is already a nice improvement.

llewelld:
A before/after preview, since I had one to hand.

flattened-elements

PR #149, opened 2022-11-23
[sailfishos][gecko] Add support for prefers-color-scheme. JB#58394

llewelld:
Adds support to qt/nsLookAndFeel for the prefers-color-scheme media tag. The renderer already supports it by setting the ui.systemUsesDarkTheme preference to 0 or 1.

This patch allows, in addition, for the colour scheme to be controlled by sending an "ambience-theme-changed" notification with a data payload of either "light" or "dark". If the ui.systemUsesDarkTheme preference is unset or set to 2, the notification value will be used to control the theme instead.

This functionality is used to update the colour scheme based on the ambience of the device.

### DIFF
--- a/rpm/0093-sailfishos-gecko-Add-support-for-prefers-color-schem.patch
+++ b/rpm/0093-sailfishos-gecko-Add-support-for-prefers-color-schem.patch
@@ -1,0 +1,154 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: David Llewellyn-Jones <david.llewellyn-jones@jolla.com>
+Date: Wed, 23 Nov 2022 09:08:22 +0200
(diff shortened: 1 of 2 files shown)


llewelld:
Part of a quiver of four PRs:
1. https://github.com/sailfishos/gecko-dev/pull/149
2. https://github.com/sailfishos/qtmozembed/pull/47
3. https://github.com/sailfishos/sailfish-browser/pull/1007
4. https://github.com/sailfishos/sailfish-components-webview/pull/161

I've added this as a separate patch, but all of the changes happen inside qt/nsLookAndFeel.cpp and qt/nsLookAndFeel.h, so this could be combined with 0002-sailfishos-qt-Bring-back-Qt-layer.-JB-50505.patch to make a single patch for both, if that's preferred.

llewelld:
Fixed a typo in the commit message.

rainemak (approved):

PR #159, opened 2024-03-10
Make header path symlink work with all gcc versions

mlehtima:


rpm/xulrunner-qt5.spec
```diff
@@ -364,7 +364,7 @@ ln -sf "%BUILD_DIR"/config.status $PWD/build/config.status
 
 %ifarch %arm32 %arm64
 # Make stdc++ headers available on a fresh path to work around include_next bug JB#55058
-if [ ! -L "%BUILD_DIR"/include ] ; then ln -s /usr/include/c++/8.3.0/ "%BUILD_DIR"/include; fi
+if [ ! -L "%BUILD_DIR"/include ] ; then ln -s /usr/include/c++/*/ "%BUILD_DIR"/include; fi
```
pvuorela, line 367: Suppose this does the trick, but pondering should there be something guarding that only a single match gets used. If there happened to be multiple ones the ln command will get interesting parameters.

Thaodan: Usually these kind of things should be done like this:
```bash
for dir in foo/* ; do
# Only take the first dir found
ln -s dir bar
break
done
```

abranson: Taking the first one means it would be likely taking the headers for the oldest gcc installed, which probably won't produce a sane build? Taking the last one would be better, though they would have to be numerically ordered properly to make sure that's from the latest version installed.

I think the most sensible thing to do if there is more than one match is to deliberately fail the build giving that as the reason.

Thaodan: > Taking the first one means it would be likely taking the headers for the oldest gcc installed, which probably won't produce a sane build? Taking the last one would be better, though they would have to be numerically ordered properly to make sure that's from the latest version installed.
> 

Than skip the break statement.

> I think the most sensible thing to do if there is more than one match is to deliberately fail the build giving that as the reason.

Technically it's not an error. The directory contains a subdirectory per compiler installed.

rainemak: > I don't think it could dependably decide the correct one to use, so I think it's fine to error out so that the problem can be noticed and dealt with. Having two compilers installed isn't a situation we need to support here.

Maybe before this block just have failure in case "ls -1 /usr/include/c++/ | wc -l" is greater than 1.

rainemak (approved):

PR #170, opened 2025-02-06
Fix searching from address bar. Re-enable multi-threaded rust builds.

mlehtima:


rpm/0090-sailfishos-gecko-Update-content-signature-root-hash..patch
```diff
@@ -0,0 +1,24 @@
...
+diff --git a/modules/libpref/init/all.js b/modules/libpref/init/all.js
+index b90f30dcffd581dc48b3fa9a824d5d3a2e2f1de8..40b180c60863fa223676c344e6a00e9d257e285e 100644
+--- a/modules/libpref/init/all.js
++++ b/modules/libpref/init/all.js
+@@ -56,7 +56,7 @@ pref("security.ssl3.rsa_aes_256_gcm_sha384", true);
+ pref("security.ssl3.rsa_des_ede3_sha", true);
+ 
+ pref("security.content.signature.root_hash",
+-     "97:E8:BA:9C:F1:2F:B3:DE:53:CC:42:A4:E6:57:7E:D6:4D:F4:93:C2:47:B4:14:FE:A0:36:81:8D:38:23:56:0E");
++     "C8:A8:0E:9A:FA:EF:4E:21:9B:6F:B5:D7:A7:1D:0F:10:12:23:BA:C5:00:1A:C2:8F:9B:0D:43:DC:59:A1:06:DB");
```
rainemak, line 21: We could live without patch by adding this to embedding/embedlite/embedding.js instead.

mlehtima: The way to add the change embedding.js would also work. The entry in embedding.js would have be removed later when updating gecko to new enough version to not have unnecessary override which could then break in the future. It's a matter of opinion which way is better.

rainemak: I think this works quite nicely like this. Like you said, not to have accidental overrides of the security.content.signature.root_hash in future.

rainemak (approved):

PR #180, opened 2026-06-16
[fix] Add Gecko shutdown phase patch. Fixes JB#64487

abranson:
Advance the EmbedLite shutdown path through the normal Gecko app/profile shutdown phases before NS_ShutdownXPCOM(). This lets profile-aware services release their worker threads and fixes qmozembed FULL_SHUTDOWN hangs.


Thaodan (approved):
LGTM.

Works well, fixes the crashes.

PR #183, opened 2026-06-19
Load XPCOM-dependent libraries with RTLD_LOCAL

abranson:
Keep Gecko's bundled mozsqlite symbols out of the process-global namespace when WebView applications also use sqlite.


abranson:
Still manages to sneak through. Think I'll have to switch back to system-sqlite.

